PR #675, opened 2023-02-28
BugFix: PCombobox null option support

pleek91:
# Description
PCombobox wouldn't show the selected value if the value was null and there was an option with a null value. This was due to some falsey checks that needed to be strict.

### DIFF
--- a/src/components/Combobox/PCombobox.vue
+++ b/src/components/Combobox/PCombobox.vue
@@ -147,11 +147,11 @@
   function displayValue(value: SelectModelValue): string {
     const option = getSelectOption(value)
 
-    if (option?.value) {
+    if (option) {
       return option.label
     }
 
-    if (!value) {
+    if (value === null) {
       return ''
     }
 


src/components/Combobox/PCombobox.vue
```diff
@@ -147,11 +147,11 @@
   function displayValue(value: SelectModelValue): string {
     const option = getSelectOption(value)
 
-    if (option?.value) {
+    if (option) {
       return option.label
     }
 
-    if (!value) {
+    if (value === null) {
```
znicholasbrown, line 154: Do we need to use a nullish check instead? `value === null || value === undefined`?

znicholasbrown (approved):
One question @pleek91

src/components/Combobox/PCombobox.vue
pleek91: That's possible but I don't think so. Based on the types
image

znicholasbrown (approved):

src/components/Combobox/PCombobox.vue
pleek91: I tested that a modelValue of `undefined` works as expected. So think the types are correct.

znicholasbrown: I appreciate the follow-up!

PR #680, opened 2023-03-08
Feature: PHeading component

pleek91:
# Description
A simple heading component
image

### DIFF
--- a/src/components/HashLink/PHashLink.vue
+++ b/src/components/HashLink/PHashLink.vue
@@ -40,6 +40,7 @@
   cursor-pointer
   relative
   transition-all
+  flex
 }
 
 .p-hash-link:before {
--- a/demo/components/ComponentPage.vue
+++ b/demo/components/ComponentPage.vue
@@ -2,7 +2,7 @@
   <div class="component-page">
     <div class="component-page__heading">
       <p-hash-link class="component-page__heading-link" :hash="title">
-        {{ title }}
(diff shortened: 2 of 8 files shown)


src/components/HashLink/PHashLink.vue
```diff
@@ -40,6 +40,7 @@
   cursor-pointer
   relative
   transition-all
+  flex
```
znicholasbrown, line 43: What's the reason for this change?

pleek91: So you can put a `p-heading` inside of a hashlink. Otherwise the hash and the text are on separate lines. This going to cause issues? I can revert. Only reason I added this was to use the p-heading component on the demo page

znicholasbrown: Ahh - this may need to be updated in a few places then, the change breaks this section of the markdown renderer and probably elsewhere that I haven't thought to check:
Screen Shot 2023-03-08 at 12 15 20 PM

znicholasbrown (approved):

src/components/HashLink/PHashLink.vue
pleek91: ahh good catch. Pushed a fix for that in the markdown parser. I cannot find any other uses of PHashLink outside of the demo in any of our projects.

brandonreid:
Nice!

PR #689, opened 2023-03-16
Feature: p-link disabled prop

stackoverfloweth:
Added prop for disabled, works for all permutations of p-link

Updated demo to include demo for disabled and for callback version of p-link

image

### DIFF
--- a/demo/sections/components/Links.vue
+++ b/demo/sections/components/Links.vue
@@ -4,6 +4,8 @@
     :demos="[
       { title: 'Local' },
       { title: 'External' },
+      { title: 'Callback' },
+      { title: 'Disabled' },
     ]"
   >
     <template #description>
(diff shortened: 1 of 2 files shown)


pleek91:
What's the motivation here? My initial thought is that we shouldn't do this.

stackoverfloweth:
@pleek91 curious why you don't think we should allow for disabling a p-link like we can a p-button? 

Especially since I can do

```html
<p-button inset :to="someroute()" disabled>
```

and end up with basically the same experience as 

```html
<p-link :to="someroute()" disabled>
```

pleek91:
Buttons natively are able to be disabled. Links are not. A disabled link is just text basically. It's also kinda difficult to disable a link in an accessible and semantic way. Taking into account screen readers and keyboard navigation. At the very least this will need `tabindex="-1"` and probably a `area-role` change?

stackoverfloweth:
@pleek91 that makes sense, I'll just close it

PR #690, opened 2023-03-16
Feature: p-modal auto-close

stackoverfloweth:
added `auto-close` prop behavior (closes modal on mask click) similar to how p-pop-over auto-close prop works

also adds prop `prevent-focus`, which let's the developer prevent automatic focusing of first focusable element on modal opening.

https://user-images.githubusercontent.com/6098901/225753876-aad23e0f-4ab2-4989-b281-60832ac432b1.mov

### DIFF
--- a/demo/sections/components/Modals.vue
+++ b/demo/sections/components/Modals.vue
@@ -1,5 +1,5 @@
 <template>
-  <ComponentPage title="Modals" :demos="[{ title: 'Modals' }]">
+  <ComponentPage title="Modals" :demos="[{ title: 'Modals' }, { title: 'Auto-Close' }]">
     <template #description>
       This is where we add a short description of <p-code inline>
         p-modals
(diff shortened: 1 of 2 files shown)


pleek91:
Whats the motivation behind `preventFocus`?

stackoverfloweth:
@pleek91 I think that functionality makes some sense when your content is like a form and the first focusable is an input but when your content doesn't have an input but does have a button or something it's pretty weird.

Automatically focusing the first focusable feels like something we threw in to solve a problem, I don't know that it's obvious behavior that developers would expect from a modal.

pleek91:
@stackoverfloweth what's wrong with focusing the button? I actually think this should go a step further and trap focus within the modal. There's a usability issue where a user can tab out of the modal and interact with the page behind the modal. 

The focus first is an incomplete solution to allow for keyboard navigation of a modal. Which I think is important and also should be consistent.

pleek91:
@stackoverfloweth I'd be in favor of trapping but not focusing automatically if that is a good compromise here.

stackoverfloweth:
@pleek91 I removed the auto-focus piece, that can be done separately

pleek91 (approved):

znicholasbrown (approved):
LGTM, thanks @stackoverfloweth 👍🏻

PR #691, opened 2023-03-17
BugFix: sizing of p-calendar date rows

stackoverfloweth:
Not sure when it happened, but I found this weird sizing bug on safari

https://user-images.githubusercontent.com/6098901/225976162-73e25deb-93bb-49ad-b988-38fdad05ea65.mov

I didn't see any recent change that explain it, but the solution I came up with is explicit row sizing for the grid on p-calendar.
A side-effect of that decision (could be pro, could be con) is that the size of each row is the same regardless of if the month has 4, 5 or 6 weeks.

https://user-images.githubusercontent.com/6098901/225976425-3e8e97e0-5e5a-42a4-ae0b-b7be3f50edc8.mov

Here's a before/after for chrome, you can see the size of content changes depending on # of weeks in viewing month

https://user-images.githubusercontent.com/6098901/225976491-246f973b-68fe-4dc2-9a69-a988cd3fba25.mov

https://user-images.githubusercontent.com/6098901/225976675-3168b351-f37c-4adc-8ae9-b3012a89d603.mov

### DIFF
--- a/src/components/Calendar/PCalendar.vue
+++ b/src/components/Calendar/PCalendar.vue
@@ -68,6 +68,7 @@
 .p-calendar__dates { @apply
   grid
   grid-cols-7
+  grid-rows-6
   gap-y-2
   gap-x-3
   flex-grow
--- a/src/components/Calendar/PCalendar.vue
+++ b/src/components/Calendar/PCalendar.vue
@@ -68,10 +68,10 @@
 .p-calendar__dates { @apply
   grid
   grid-cols-7
-  grid-rows-6
   gap-y-2
   gap-x-3
-  flex-grow
+  flex-grow;
+  grid-auto-rows: minmax(0, 1fr)
 }
 
 .p-calendar__date { @apply


src/components/Calendar/PCalendar.vue
```diff
@@ -68,6 +68,7 @@
 .p-calendar__dates { @apply
   grid
   grid-cols-7
+  grid-rows-6
```
pleek91, line 71: so this tailwind class makes the css `grid-template-rows: repeat(6, minmax(0, 1fr))`. I believe the `minmax` is the important part and not the 6. I think you can use `grid-auto-rows: minmax(0, 1fr)`

pleek91 (commented):
I was unable to reproduce in safari. But I think this is a good change still. One suggestion that I think will fix the number of weeks issue.

src/components/Calendar/PCalendar.vue
stackoverfloweth: @pleek91 that is true, that means rows fill the space available, assuming we prefer that over set height regardless of viewing month?

pleek91: I think so? That leaves the existing layout but fixes it for safari right? Again, I'm unable to reproduce.

pleek91: @stackoverfloweth

stackoverfloweth: applied suggestion, thanks

pleek91 (approved):

PR #692, opened 2023-03-17
Feature: p-native-datepicker new show-time prop

stackoverfloweth:
as you probably know, p-date-input has a prop for `show-time`, which lets the user set both the date and the time.
as you probably also know, p-date-input relies on p-native-date-input if `media.hover`, which ignores `show-time`.

https://user-images.githubusercontent.com/6098901/225978874-d86ec99e-c210-4983-9d46-730a59dc2f5b.mov

https://user-images.githubusercontent.com/6098901/225979201-e7e3df13-5830-4bff-a159-16889ef2311b.MP4

my solution to add support for `show-time` prop on p-native-date-input, which puts an icon in the input that toggles between entering the "date" portion and entering the "time" portion.

https://user-images.githubusercontent.com/6098901/225978965-fa77245d-e99f-4f80-9c98-00a370005d62.mov

https://user-images.githubusercontent.com/6098901/225978988-b708bdfd-2a9b-4a33-80a0-a5676f6c7c5d.MP4

### DIFF
--- a/src/components/DateInput/PDateInput.vue
+++ b/src/components/DateInput/PDateInput.vue
@@ -21,7 +21,7 @@
           <PNativeDateInput
             v-model="internalModelValue"
             class="p-date-input__native"
-            v-bind="{ min, max, disabled }"
+            v-bind="{ min, max, disabled, showTime }"
           />
         </template>
       </slot>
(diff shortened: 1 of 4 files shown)


pleek91 (approved):
I personally don't love the icon approach to switch but this is 100% better than what we have now. Love the time parts abstraction. Thanks @stackoverfloweth

PR #693, opened 2023-03-17
Feature: expose searchValue on p-combobox

stackoverfloweth:
prop and events that expose previously internal `typedValue` (now called `searchValue`)

```html
<p-combobox v-model="value" :option="options" @update:searchValue="respondToSearch" />
```

developer could also theoretically tie the value to another value or pre-populate the search with `v-model:searchValue`

planning to use this to fetch options async as user types, think of like an address input that brings in results from google maps as you type your address

### DIFF
--- a/src/components/Combobox/PCombobox.vue
+++ b/src/components/Combobox/PCombobox.vue
@@ -17,7 +17,7 @@
       <div class="p-combobox__search-option">
         <input
           ref="textInput"
-          v-model="searchValue"
+          v-model="search"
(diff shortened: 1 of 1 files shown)


pleek91:
@stackoverfloweth this is a nice enhancement. Do you feel strongly about `searchValue` as a name? `value` is implied IMO and doesn't add much. I'd suggest just `search` so the syntax is `v-model:search`.

pleek91 (approved):

PR #698, opened 2023-03-21
BugFix: PDivider not very visible

pleek91:
# Description
Matching the colors from PCard

Before
image

After
image

### DIFF
--- a/src/components/Divider/PDivider.vue
+++ b/src/components/Divider/PDivider.vue
@@ -7,7 +7,7 @@
     border-none
     h-[1px]
     bg-background-400
-    dark:bg-background-500
+    dark:bg-foreground-200
     w-full
     my-3
   }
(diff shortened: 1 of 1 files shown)


znicholasbrown:
Hm, this was intentional tbh, it's meant to be lower-contrast than content as a visual rest but not a hard break necessarily. That said, I don't feel strongly about the change (except the change to `foreground`, that makes sense)

pleek91:
Its just not even noticeable in some contexts currently which is what I'm trying to solve. And if it matches the card in light mode I think it should match the card in dark mode as well.

znicholasbrown:
Hm ok I see your point; I don't think the card border is the correct standard here since this is for content. Tested with `bg-foreground-50` and `dark:bg-foreground-100` and thought the results were decent, wdyt?

pleek91:
Looks good on light mode but personally I can barely even see the divider on dark mode. Like I cannot see if except that I'm looking for it

image

pleek91:
@znicholasbrown never mind. That screenshot is incorrect. Yes those colors look good to me. Updated 👍

znicholasbrown (approved):
ty!

PR #701, opened 2023-03-23
Enhancement: PCard flat prop + animation

znicholasbrown:
Pretty straightforward enhancement to the `PCard` component that adds a boolean `flat` property that removes the component's border, background, and box-shadow but keeps the padding. It also transitions all of these except box shadow.

### DIFF
--- a/src/components/Card/PCard.vue
+++ b/src/components/Card/PCard.vue
@@ -8,7 +8,7 @@
   import { computed } from 'vue'
 
   const props = defineProps<{
-    flat: boolean,
+    flat?: boolean,
   }>()
 
   const classes = computed(() => ({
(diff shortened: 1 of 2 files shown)


src/components/Card/PCard.vue
```diff
@@ -14,5 +26,20 @@
...
   rounded-lg
   shadow-sm
+  transition-all;
+
+  /*
+    Explicitly defining these after transition-all
+      to override the transition-property on the box-shadow
+      because transitioning shadows isn't super performant
+      and it doesn't add much value.
+  */
```
pleek91, line 36: Personally I don't think the comment is needed. We do this in a few places where we have tailwind and then plain css overrides or extra stuff.

src/components/Card/PCard.vue
```diff
@@ -14,5 +26,20 @@
...
+  transition-property: border border-color border-radius background-color color padding;
```
pleek91, line 37: Why color padding and radius? I don't think your changing those properties.

pleek91 (approved):

PR #704, opened 2023-03-25
BugFix: make sure bg can scroll after modal is closed

stackoverfloweth:
looks like original implementation mistook the 2nd argument as boolean for "should the class be there or not" when in reality the 2nd argument is a boolean for "force". 
https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList/toggle

also ensure bg-scroll gets re-enabled if p-modal gets unmounted without resetting `showModal`

### DIFF
--- a/src/components/Modal/PModal.vue
+++ b/src/components/Modal/PModal.vue
@@ -122,10 +122,6 @@
     }
   }
 
-  function disableBackgroundScroll(): void {
-    document.body.classList.add('p-modal__stop-bg-scroll')
-  }
-
   function enableBackgroundScroll(): void {
     document.body.classList.remove('p-modal__stop-bg-scroll')
   }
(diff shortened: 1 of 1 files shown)


pleek91:
The original implementation looks correct to me? If showModal is true then the force argument is true which forces the class to be added. If showModal is false then the false argument is false forcing the class to be removed. I've used that pattern dozens of times, pretty sure it's correct.

stackoverfloweth:
@pleek91 interesting, not what I expected by a param "force".
Do you think there is value in the the onBeforeUnmount piece?

pleek91:
@stackoverfloweth seems like a logical safeguard though I've never run into that scenario personally

stackoverfloweth:
@pleek91 I reverted the classList toggle piece, up to you if you wanna merge or close for the onBeforeUnmount piece. I ran into it because I had a `v-if` around my `p-modal`

pleek91 (approved):

PR #705, opened 2023-03-26
Chore: List item and list item input BEM classes

stackoverfloweth:
now developers can have a @click event on <p-list-item-input /> that will not get triggered by selection change

this PR also fixes some incorrect BEM classes on `p-list-item` and `p-list-item-input`

### DIFF
--- a/src/components/ListItem/PListItem.vue
+++ b/src/components/ListItem/PListItem.vue
@@ -1,11 +1,11 @@
 <template>
-  <div class="list-item">
+  <div class="p-list-item">
     <slot />
   </div>
 </template>
 
 <style>
-.list-item { @apply
+.p-list-item { @apply
   bg-background
   shadow
   rounded-lg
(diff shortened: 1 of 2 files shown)


pleek91:
@stackoverfloweth this is so you could have the whole list item have a click action? That feels like a "link inside a link" which isn't a great UX or devx. 

I'm thinking a click.stop also means that any global click handlers (for closing an overflow menu for example) won't be triggered.

stackoverfloweth:
@pleek91 I don't disagree, thoughts on just emitting a separate event for clicking on content vs clicking on checkbox? (see last commit)

pleek91:
Not sure what side effects having a custom event with the same name as a native event. 

Also thinking, couldn't this be an implementation detail by adding a click handler to the slots content itself?

stackoverfloweth:
@pleek91 good call, I like that even better. I'll leave only the BEM fixes here if you wanna make that change

pleek91 (approved):

PR #709, opened 2023-03-28
Chore: Fix extraneous non-props warning in PTabs

pleek91:
# Description
The class that was added to PTabSelect was causing a warning because that component uses fragments. Seems like it shouldn't be a fragment component but rather than refactor I'm removing the unused class that was causing the warning. 

image

### DIFF
--- a/src/components/Tabs/PTabs.vue
+++ b/src/components/Tabs/PTabs.vue
@@ -8,7 +8,7 @@
       </PTabNavigation>
     </template>
     <template v-else>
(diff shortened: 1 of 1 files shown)


znicholasbrown (approved):
This feels like a breaking change but tabs look good (mobile and otherwise) and I don't see this class on the component so LGTM

pleek91:
Technically if anyone is targeting that class specifically this would break something. But we're not anywhere so 🤷

PR #712, opened 2023-03-30
Bugfix: Add timeline types to bucket

znicholasbrown:


pleek91 (approved):
Forever going to call these buckets ❤️

PR #713, opened 2023-03-30
Enhancement: Add `items-start` to `.p-key-value`

pleek91:
# Description
`p-key-value` is `flex` and `flex-col`. Which means by default the flex children expand to fill the width. This means that the rect of the label and value are displayed like block elements. This can cause issues when the content of these slots has hover or click events because those events will be triggered in the empty space. 

Side quest: I also removed the duplicate styles from the `--alt` class that were already specified in the block class. 

Before
image

After
image

### DIFF
--- a/src/components/KeyValue/PKeyValue.vue
+++ b/src/components/KeyValue/PKeyValue.vue
@@ -73,6 +73,7 @@
   flex-col
   gap-y-1
   leading-6
+  items-start
 }
 
 .p-key-value__label {
@@ -94,9 +95,6 @@
   @apply
   text-xs
   leading-4
-  flex
-  flex-col
-  gap-y-1
   font-normal
 }
 


znicholasbrown (approved):
Nice, makes sense

znicholasbrown:
@brandonreid I think we use them in a few places for larger content displays on mobile

brandonreid:
> @brandonreid I think we use them in a few places for larger content displays on mobile

Maybe a safer bet is `inline-flex`? That way the gap is preserved for things downstream that might use it?

pleek91:
> > @brandonreid I think we use them in a few places for larger content displays on mobile
> 
> 
> 
> Maybe a safer bet is `inline-flex`? That way the gap is preserved for things downstream that might use it?

@brandonreid are you concerned about the styles I removed from the alt class or the items-start addition? I'm not following your comments. Nothing should change for layout or gap except for the rect width of the internal elements.

src/components/KeyValue/PKeyValue.vue
```diff
@@ -94,9 +95,6 @@
   @apply
   text-xs
   leading-4
-  flex
-  flex-col
-  gap-y-1
```
brandonreid, line 99: @pleek91 I'm not overly concerned here. But these flex styles tell me that if there were multiple items within the label, they'd have column styling with a gap. Removing them would probably revert them to inline or something.

pleek91: Okay so these are part of the BEM modifier class `p-key-value--alt`. It's intent is to modify the `p-key-value` block class. But that block class already has these styles so they were being applied twice (I noticed this in the dom when testing). No functional change of styles is being made here. Because this modifier class is not used or intended to be used on its own and these styles are present in the primary block level class. 

I can go into more detail about BEM class semantics sometime if you still have questions about this

brandonreid: Ah ok. I never looked at the root class.

No need for a BEM review, haha, even though I have strayed from the path many times before, thank you though.

PR #715, opened 2023-03-31
Feature: PTooltip component

pleek91:
# Description
Adds a new PTooltip component. Supports tooltip text as a prop as well as two slots to customize the tooltip itself. 

image


src/components/Tooltip/PTooltip.vue
```diff
@@ -0,0 +1,88 @@
...
+  import { top, bottom, left, right } from '@/utilities/position'
+
+  const props = defineProps<{
+    text?: string,
+    placement?: PositionMethod | PositionMethod[],
+    to?: string | Element,
+  }>()
+
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-type-arguments
+  const popover = ref<InstanceType<typeof PPopOver>>()
```
pleek91, line 34: I have no idea why but eslint kept removing this type which would make it `Ref<any>` which isn't helpful.

stackoverfloweth (approved):
Can't believe we were able to avoid this component for as long as we did. This is top notch @pleek91. Unfortunate about the eslint thing but otherwise exactly what I expected.

pleek91:
Thanks @stackoverfloweth!

src/components/Tooltip/PTooltip.vue
brandonreid: I used an a similar type and needed to instantiate it with `null`. So like `ref<InstanceType<typeof PPopOver> | null>(null)`, maybe the popover doesn't exist until some mounting event so it's not possible the const assignment is valid on instantiation?

pleek91: That did work. Thanks! Still doesn't make sense though. The ref doesn't exist until mounting but the type of a ref without a value is `ref<T | undefined>`. But least `null` gets rid of the linting issue.

pleek91:
@brandonreid and @stackoverfloweth I added a couple more commits to this after looking at [ARIA: tooltip role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tooltip_role) from MDN. 
- Added a 500ms delay on opening
- Added the ability to press escape to close the tooltip
- Added the ability to open the tooltip on focus (keyboard navigation)
- Added `role` and `aria-describedby` attributes to appropriate elements.

brandonreid (approved):

PR #721, opened 2023-04-06
Enhancement: Add margin to timeline date when stacked

pleek91:
# Description
The date and content had no gap when stacked. Adds some margin to the date to be consistent with the `--p-timeline-item-gap` custom property

### DIFF
--- a/src/components/Timeline/PTimelineItem.vue
+++ b/src/components/Timeline/PTimelineItem.vue
@@ -101,6 +101,10 @@
   min-w-0
 }
 
+.p-timeline-item__date:empty { @apply
+  hidden
+}
+
 .p-timeline-item__date {
   grid-area: date;
 }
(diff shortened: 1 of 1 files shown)


stackoverfloweth (approved):

znicholasbrown (changes requested):
Hm wait this was intentional and adding a gap leads to this (note the weird spacing when that slot isn't populated):

image

vs main:

Screenshot 2023-04-06 at 2 07 21 PM

We either need to account for that by removing the slot wrapper if it's not provided or teh caller should be responsible for adding a margin

pleek91:
@znicholasbrown good catch. There was already a `v-if` on the date slot but I forgot that the slot can exist but still be empty. Added a css `:empty` so that even if the slot exists, if it doesn't return anything it'll still be hidden.

znicholasbrown (approved):
Nice solution, thanks @pleek91 👍🏻

PR #722, opened 2023-04-06
Bugfix: Remove overscroll-contain on PCodeInput

znicholasbrown:
Contained overscroll is only really nice for small inputs; large inputs capture too much passive scrolling and lead to a poor UX if they don't pass the event up the chain.


pleek91 (approved):
Yeah this is probably the best default

PR #727, opened 2023-04-13
BugFix: background on p-tooltip

stackoverfloweth:
simply using the non-theme aware color

before: (only noticeable on light-mode)
Screenshot 2023-04-13 at 9 02 40 AM

after:
Screenshot 2023-04-13 at 9 02 52 AM
Screenshot 2023-04-13 at 9 02 59 AM

### DIFF
--- a/src/components/Tooltip/PTooltip.vue
+++ b/src/components/Tooltip/PTooltip.vue
@@ -99,6 +99,7 @@
   text-sm
   rounded-sm
   border-[1px]
-  border-slate-600
+  border-background-400
+  dark:border-foreground-200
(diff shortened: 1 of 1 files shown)


brandonreid (approved):
🙌

pleek91:
Dude thank you! While you're at it, do you mind updating the borders for the tooltip? Using `bg-background` rather than slate is perfect. But now the border seems a little off especially on light mode. I think we should match p-card here

```
  border-background-400
  dark:border-foreground-200
```

pleek91 (approved):

PR #728, opened 2023-04-13
Enhancement: Increase PButton inset style contrast in dark mode

pleek91:
# Description
On dark mode the border of the "inset" style is too dark and doesn't provide enough contrast. Making the border lighter for dark mode make it much more distinguishable. 

Before
image
image
image
image

After
image
image
image
image


znicholasbrown (approved):

zhen0 (approved):
Thanks @pleek91

PR #732, opened 2023-04-19
BugFix: PTabs key bound to tab object

pleek91:
# Description
When rendering tabs the entire tab object is being used as the `key`. This means tabs are not reused if the tabs array is updated. Switching the key to be the tab label to enable vue to accurately track tabs.


znicholasbrown (approved):
thanks @pleek91

PR #733, opened 2023-04-19
reduce the size of pTab

brandonreid:
The tabs on the Flow Run Page are falling off the viewport.

Currently:
image

This shrinks them down so they look like this:

https://user-images.githubusercontent.com/6776415/233161085-82acc9d8-522c-47a0-92f7-af616ae6e185.mov

### DIFF
--- a/src/components/Tab/PTab.vue
+++ b/src/components/Tab/PTab.vue
@@ -20,7 +20,7 @@
 
 <style>
 .p-tab { @apply
-  px-10
+  px-5
   py-4
   text-center
   font-medium


znicholasbrown:
@brandonreid this doesn't feel like a sustainable solution to overflowing tabs; it fixes the issue on the specific screen size in your video but not on any smaller ones

brandonreid:
@znicholasbrown did you play the video by chance? I shrink the viewport and show that it keeps everything visible and without horizontal scroll down to mobile.

I anticipate the number of tabs shrinking once filtering/search is available on the timeline (will remove Sub Flow Runs and Task Runs). Certainly room to get more hard core with it when we have time.

pleek91:
> @brandonreid this doesn't feel like a sustainable solution to overflowing tabs; it fixes the issue on the specific screen size in your video but not on any smaller ones

+1 but I also think the smaller tabs look better. So I'm fine with approving this even knowing it doesn't fix the actual problem of PTabs not being responsive.

znicholasbrown:
Sorry @brandonreid I should have been more specific with that feedback - this is a stopgap solution to fix a specific use of `PTabs` but does nothing to address the actual issue. Visually it looks nice but even in your video there's a screen range where this is broken (at the 5 second mark). We can accept this but I'd really like to see a more thoughtful solution

brandonreid:
@znicholasbrown yeah I gathered that after. Agree for sure.

PR #748, opened 2023-05-04
Feature: `PLayoutResizable`

znicholasbrown:
This PR introduces a new layout, `PLayoutResizable`, that encapsulates all the logic needed to create a layout like the following:
```
|           |                  |
|   aside   |       main       |
|           |                  |
```
where the middle bar lets you change the dimensions of the two sides.

This is helpful in app layouts with a global sidebar that can be expanded/shrunk. It also includes functionality for a "collapse point" the aside can automatically shrink to, enabling a nice collapsible feel. 

```
|     |                      |
|  a  |         main         |
|     |                      |
```


demo/sections/layouts/LayoutSidebar.vue
znicholasbrown, line 1: I removed this demo since it wasn't actually showcasing a component and has been broken for awhile

src/layouts/PLayoutResizable/PLayoutResizable.vue
```diff
@@ -0,0 +1,174 @@
...
+
+<style>
+.p-layout-resizable {
+  --aside-width: var(--p-layout-resizable-aside-width, min-content);
+  --aside-max-width: var(--p-layout-resizable-aside-max-width, theme(width.80));
+  --aside-min-width: var(--p-layout-resizable-aside-min-width, theme(width.20));
+}
+
+.p-layout-resizable { @apply
+  flex
```
pleek91, line 113: I'm a little concerned about using flex here. I've just been burned by flex "blowout" a lot. But I'm also not sure if you could do this with grid. 

I think at the very least we might need a `min-width: 0` on the `__main` element.

src/layouts/PLayoutResizable/PLayoutResizable.vue
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  const start = (): void => {
+    if (props.disabled || !container.value || !aside.value) {
+      return
+    }
+
+    const asideBounds = aside.value.getBoundingClientRect()
+    asideLeft = asideBounds.left
+    document.body.style.cursor = 'col-resize'
```
pleek91, line 71: manipulating this directly seems a little sketchy. Can we make a class and toggle the class on the body?

src/layouts/PLayoutResizable/PLayoutResizable.vue
```diff
@@ -0,0 +1,174 @@
...
+    dragging.value = true
+    document.body.style.userSelect = 'none'
```
pleek91, line 73: Same with this

src/layouts/PLayoutResizable/PLayoutResizable.vue
znicholasbrown: Yeah makes sense - I added that. If that becomes a problem I'll revisit this but for now it works nicely without

src/layouts/PLayoutResizable/PLayoutResizable.vue
znicholasbrown: Lol you're completely right, idk why I went with that - added a class method instead.

pleek91 (approved):

PR #769, opened 2023-05-18
Enhancement: Update PTable types and use generics

pleek91:
# Description
Updating to latest vue-tsc meant slot types were generated automatically. This raised a lot of upstream errors in implementations that were written based on the slot scope being type `any`. Using generics and improving the types in general for p-table makes the migration much easier.


pleek91:
@znicholasbrown if you have time to look at this I'd appreciate it.

src/components/Table/PTable.vue
```diff
@@ -57,7 +57,7 @@
   </div>
 </template>
 
-<script lang="ts" setup>
+<script lang="ts" generic="const TData extends TableData, const TColumn extends TableColumn<TData>" setup>
```
znicholasbrown, line 60: ooooo

src/components/Table/PTable.vue
```diff
@@ -138,19 +134,19 @@
...
-  function convertStringsToTableColumns(columns: string[]): TableColumn[] {
-    return columns.map(property => ({ label: property, property }))
+  function convertStringsToTableColumns(columns: string[]): TColumn[] {
+    return columns.map(property => ({ label: property, property })) as TColumn[]
   }
 
-  function getValue(row: TableData, column: TableColumn): unknown {
-    if (column.property) {
-      return row[column.property]
+  function getValue<P extends keyof TData>(row: TData, property: P | undefined): any {
```
znicholasbrown, line 141: Warning here - is `any` correct? I'd expect to return a generic here

znicholasbrown (approved):
Just one question, maybe requiring an any ignore comment - this syntax is new to me so I can't pretend I know if this will cause issues but it aligns with what I'd expect

src/components/Table/PTable.vue
pleek91: ah I should have explained the reason for the `any`. I added an eslint-ignore and a comment for this. But basically this is for backwards compatibility. A column slot is based on the column label. Which does not have the restriction of being a `keyof TData`. It can be any string. If you add an optional `property: keyof TData` then it will auto populate the value and bind it to the slot scope. I feel like it might be possible to infer the type based on the `property` if there is one and hard code it to `undefined` if there is not. But that would take some typescript magic that I'm not prepared to wrangle right now lol.

PR #775, opened 2023-05-22
Enhancement: Use generics for p-virtual-scroller

pleek91:
# Description
Previously the slot scope type for `item` was hard coded to `any`. With generics we can make that inferred.


znicholasbrown (approved):
yessss

Issue #774, opened 2023-05-22
Indicate that select menus are scrollable


brandonreid:
In [this issue](https://github.com/PrefectHQ/nebula-ui/issues/2783), a user was not aware of additional options or the ability to scroll. I propose 2 options here.

## The simple/quick option

We adopt a design principle for this component around applying max heights that include a peaking option. So perhaps the component is smart enough to know the height of options and self assign a max height that is something like `true max minus half of one option`. Or we simply state that this should be the desired outcome and the implementation assigns the max height that allows for a peeking option.

## The over-engineered option

When options go higher than the max height of the container we add a scroll indicator:
<img width="1164" alt="image" src="https://github.com/PrefectHQ/prefect-design/assets/6776415/eb770d9e-8576-4a22-a0e8-895cc1b217c2">
Here you see a gradient overlay and two arrows indicating there is more to see. We'd need to clear this indicator any time a scroll event takes place. There's likely some subtlety to play with here around mouse over events, maybe the indicator slowly fades away.

PR #779, opened 2023-05-22
Enhancement: Add visual queue to p-selection options dropdown when there is overflow

pleek91:
# Description
Middle ground between the options in the following issue. Added a gradient overlay when you can scroll down. Not against adding the icon as well but IMO its a little confusing on if you can click on it or what. 
Closes https://github.com/PrefectHQ/prefect-design/issues/774

image

### DIFF
--- a/src/components/Select/PSelectOptions.vue
+++ b/src/components/Select/PSelectOptions.vue
@@ -42,7 +42,7 @@
 </template>
 
 <script lang="ts" setup>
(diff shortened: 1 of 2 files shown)


znicholasbrown (approved):
I like this as a compromise - curious for @brandonreid's thoughts

brandonreid (approved):
Nice, yeah does the job great. Def agree with not needing the arrows!

PR #799, opened 2023-05-31
Enhancement: `PTooltip` styling and width fix

znicholasbrown:
My goal with the new styles was to make the tooltips look like less cards and foreground elements we use frequently in this library (opaque backgrounds with thin, rounded borders) and a little more "ephemeral". To that end I've added distinct box shadows, a slight blur on the background, and ∫umped the font size down 1 step. 

Width fix before:
Screenshot 2023-05-31 at 3 39 44 PM

After:
Screenshot 2023-05-31 at 3 49 46 PM

Styling changes before:
Screenshot 2023-05-31 at 4 57 36 PM
Screenshot 2023-05-31 at 4 57 28 PM

After:
Screenshot 2023-05-31 at 4 57 54 PM
Screenshot 2023-05-31 at 4 57 44 PM

### DIFF
--- a/src/components/Tooltip/PTooltip.vue
+++ b/src/components/Tooltip/PTooltip.vue
@@ -18,7 +18,7 @@
 </template>
 
 <script lang="ts" setup>
-  import { computed, onMounted, onUnmounted, ref } from 'vue'
+  import { computed, onUnmounted, ref } from 'vue'
   import PPopOver from '@/components/PopOver/PPopOver.vue'
   import { PositionMethod } from '@/types/position'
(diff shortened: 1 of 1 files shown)


znicholasbrown:
Hm yeah I see your point - try now, I bumped the font size back up and made the background a little more opaque

brandonreid (approved):
Reviewed the updates in the deployment preview. As these tooltips should act as a sort of quiet helper, I really like the glass effect overall. When viewing as a dev it's easy to tunnel vision on it a bit, but in action I think it will feel nice.

PR #801, opened 2023-06-01
fix modal styles

brandonreid:
Old light mode:
image

New light mode:
image

Old dark mode:
image

New dark mode:
image


pleek91:
Thank you!

PR #809, opened 2023-06-08
Enhancement: Pop over groups

pleek91:
# Description
Introducing the concepts of groups to the popover component. Adding a group value enforces that only a single popover in that group will be open at a time. If a popover in a group is open and another popover in the same group is opened the already opened popover will close automatically.


src/components/PopOver/compositions/usePopOverGroup.ts
```diff
@@ -0,0 +1,26 @@
+export type PopOverOpenMethod = () => void
+export type PopOverCloseMethod = () => void
+
+export type UsePopOverGroup = () => {
+  setGroupCloseMethod: (groupId: string, close: PopOverCloseMethod) => void,
+}
+
+function factory(): UsePopOverGroup {
```
brandonreid, line 8: I'm curious why you chose to name this function `factory` rather than exporting `usePopOverGroup` directly?

brandonreid (approved):
Lgtm overall. In the demo, I noticed the last tooltip persists even when I hover away from the group. I'm sure that demo misses some level of implementation detail.

pleek91:
> Lgtm overall. In the demo, I noticed the last tooltip persists even when I hover away from the group. I'm sure that demo misses some level of implementation detail.

The open/close logic is always decided at implementation. For this demo its opened by a mouse over and the `auto-close` prop is added which closes the popover if you click outside of the target or the popover content.

src/components/PopOver/compositions/usePopOverGroup.ts
pleek91: This is a pattern I've used in quite a few compositions to accomplish encapsulation of a shared state. Basically write a factory method that returns the actual exported composition (bottom of the file). Which allows the composition to share some values across all composition calls (in this case the `groups` const and the `setGroupCloseMethod` function).

brandonreid: Ahh, gotchya. Rather than creating multiple instances of the function. Makes sense, thank you!

pleek91:
@brandonreid your comment about not closing when leaving the group gave me an idea. I updated the composition to expose both open and close methods and I exported the composition. This means anywhere you can call the composition and get a `close` method that acts on the group. Which allows you to easily close a popover from anywhere which we didn't have the ability to do before. Updated the demo to act in the way you mentioned.

brandonreid:
> @brandonreid your comment about not closing when leaving the group gave me an idea. I updated the composition to expose both open and close methods and I exported the composition. This means anywhere you can call the composition and get a `close` method that acts on the group. Which allows you to easily close a popover from anywhere which we didn't have the ability to do before. Updated the demo to act in the way you mentioned.

@pleek91 slick! I could see that coming in handy on a page like the dashboard where you've def got multiple groups (list of mini history graphs on the work pools cards), just to at least have a sure bet only one is ever showing.

PR #810, opened 2023-06-08
Enhancement: `PLayout*` conditional slots

znicholasbrown:
**This is a breaking change**

This PR removes the default padding on the `PLayoutDefault` and `PLayoutWell` components. It also hides the wrappers for slots that aren't present to prevent adding spacing where it shouldn't exist. 

I'm a little tentative to introduce these changes but without more flexibility here we don't have much control when using these layouts so I'm open to closing this in favor of a different solution

### DIFF
--- a/src/layouts/PLayoutDefault/PLayoutDefault.vue
+++ b/src/layouts/PLayoutDefault/PLayoutDefault.vue
@@ -1,7 +1,7 @@
 <template>
   <div class="p-layout-default">
-    <div class="p-layout-default__header">
-      <slot name="header" class="p-layout-default__header" />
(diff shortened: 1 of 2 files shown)


brandonreid (approved):

brandonreid:
I guess I see it as pretty low risk. Less _breaking_ and more _borking_, haha. Seems easy enough to revert.

pleek91:
So this just leaves the spacing around the content up to the implementation? I guess that makes sense and also makes these layouts more consistent considering the other layouts don't have padding. Fine by me 👍

pleek91 (approved):

znicholasbrown:
Closing this for now

PR #816, opened 2023-06-12
Feature: PAccordion

pleek91:
# Description
Adds a new accordion component. Takes the "at most one" strategy meaning sections will toggle automatically to keep only a single section open. But you can close the open section and have no sections open. 

image

### DIFF
--- a/src/components/Accordion/PAccordion.vue
+++ b/src/components/Accordion/PAccordion.vue
@@ -2,18 +2,27 @@
   <div class="p-accordion">
     <template v-for="section in sections" :key="section">
       <div class="p-accordion__section">
-        <div class="p-accordion__header" @click="openSection(String(section))">
+        <div class="p-accordion__header">
           <!-- dynamic slot names don't work great yet -->
           <!-- @vue-ignore -->
           <slot :name="`${section}-heading`">
             {{ section }}
           </slot>
+          <p-button
+            size="sm"
+            icon="ChevronDownIcon"
+            flat
+            rounded
+            class="p-accordion__button"
+            :class="getButtonClass(String(section))"
+            @click="toggleSection(String(section))"
+          />
         </div>
 
         <PAutoHeightTransition>
           <template v-if="isOpen(String(section))">
             <div class="p-accordion__content">
-              <div class="pb-2">
+              <div class="p-accordion__content-padding">
                 <!-- dynamic slot names don't work great yet -->
                 <!-- @vue-ignore -->
                 <slot :name="`${section}-content`" />
@@ -46,14 +55,25 @@
 
   defineSlots<AccordionSlots<T>>()
 
-  const open: Ref<T[number]> = ref(props.sections[0])
+  const open: Ref<T[number] | null> = ref(props.sections[0])
 
-  function openSection(section: string): void {
+  function toggleSection(section: string): void {
+    if (isOpen(section)) {
+      open.value = null
+      return
+    }
+
     open.value = section
   }
 
   function isOpen(section: string): boolean {
     return open.value === section
+  }
+
+  function getButtonClass(section: string): string | undefined {
+    if (isOpen(section)) {
+      return 'p-accordion__button--open'
+    }
   }
 </script>
 
@@ -63,5 +83,20 @@
   py-2
   border-t
   border-foreground-200
+  flex
+  justify-between
+  items-center
+}
+
+.p-accordion__content-padding { @apply
+  pb-2
+}
+
+.p-accordion__button { @apply
+  transition-transform
+}
+
+.p-accordion__button--open { @apply
+  rotate-180
 }
 </style>
--- a/src/components/Accordion/PAccordion.vue
+++ b/src/components/Accordion/PAccordion.vue
@@ -6,7 +6,11 @@
           <!-- dynamic slot names don't work great yet -->
           <!-- @vue-ignore -->
           <slot :name="`${section}-heading`">
-            {{ section }}
+            <!-- dynamic slot names don't work great yet -->
+            <!-- @vue-ignore -->
+            <slot name="heading" :section="section">
+              {{ section }}
+            </slot>
           </slot>
           <p-button
             size="sm"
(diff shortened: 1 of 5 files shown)


src/components/Accordion/PAccordion.vue
```diff
@@ -0,0 +1,91 @@
...
+            :class="getButtonClass(String(section))"
+            @click="toggleSection(String(section))"
+          />
+        </div>
+
+        <PAutoHeightTransition>
+          <template v-if="isOpen(String(section))">
+            <div class="p-accordion__content">
+              <!-- this is needed because PAutoHeightTransition does not animate an elements padding correctly -->
+              <div class="p-accordion__content-padding">
```
brandonreid, line 26: Maybe a `wrapper` and `content` naming convention would make this more BEMy?

src/components/Accordion/PAccordion.vue
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    open.value = section
+  }
+
+  function isOpen(section: string): boolean {
+    return open.value === section
+  }
+
+  function getButtonClass(section: string): string | undefined {
```
brandonreid, line 62: Elsewhere we compute a list of conditional classes, nbd. A little surprised this is recalled when the state changes.

src/components/Accordion/PAccordion.vue
```diff
@@ -0,0 +1,91 @@
...
+
+<script lang="ts" setup generic="const T extends Readonly<string[]>">
+  import { Ref, ref } from 'vue'
+  import { PAutoHeightTransition } from '@/components/AutoHeightTransition'
+
+  const props = defineProps<{
+    sections: T,
+  }>()
+
+  const open: Ref<T[number] | null> = ref(props.sections[0])
```
brandonreid, line 47: So the first section is always open by default? I may end up updating to allow for a default open section with the onboarding module, but we'll see when we get there.

brandonreid (approved):
Couple nits, nbd.

I do think we're continuing our practice of skipping accessibility, like we could add the [recommended decorators](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/examples/accordion/) and such (we'd have to group the header and button differently in ours of course). Not sure how we feel on that, from what I know we haven't heard any issues and even this pattern, it could be used even if it's not optimized for the screen reader.

src/components/Accordion/PAccordion.vue
znicholasbrown, line 47: Would like this to be configurable instead of always grabbing the first one - the type for this is also slightly incorrect since an empty array fulfills `string[]` and so will return `undefined` in this instance.

src/components/Accordion/PAccordion.vue
```diff
@@ -0,0 +1,91 @@
...
+              {{ section }}
+            </slot>
+          </slot>
+          <p-button
+            size="sm"
+            icon="ChevronDownIcon"
+            flat
+            rounded
+            class="p-accordion__button"
+            :class="getButtonClass(String(section))"
```
znicholasbrown, line 17: I see you casting the `section` to string in quite a few places here - can you explain why that's necessary? I see that these methods expect a string but I'd expect an iterated `Readonly<string[]>` to leave you with a `string`

znicholasbrown (commented):
tldr; the component is a little confusing but that might be by design. Non-blocking comments only.

I might be wrong here but my expectation for accordion items is that the entire heading acts as the button and the chevron is mainly an indicator that it's interactive. My first instinct when interacting with the demo component was to click a bunch on the header and its content which led to a (mildly) frustrating experience.

Separate from the interaction design, the visual design doesn't give any visual cues to differentiate section heading from content. This sort of makes sense in the current interaction pattern where only the chevron is clickable but isn't a terrific experience overall.

Lastly, and to @brandonreid's comments, I think we're missing some helpful aria attrs and semantic HTML.

src/components/Accordion/PAccordion.vue
pleek91: I'll make this a v-model so it can be optionally customized by the implementation.

src/components/Accordion/PAccordion.vue
pleek91: Yeah frequently you'll see a computed for something like this. But when you're getting dynamic classes for elements in a v-for this is usually how that's done. Vue is pretty smart about re-evaluating method calls that are referenced in the template and I've never had an issue with this. Computes are preferred but in a v-for a method is just much simpler. Otherwise we'd have to make a computed with dynamic keys or something.

src/components/Accordion/PAccordion.vue
pleek91: Both follow BEM. That second div is only necessary because of the transition bug. So I think an overly specific class helps keep that div from being used for anything else so that it can hopefully be removed in the future.

pleek91:
@znicholasbrown I also prefer the whole heading be clickable. But the design I'm working off of has pieces of the heading as links. And I want to avoid the "link inside a button" ux issue. @brandonreid maybe you have some thoughts on this or maybe I'm just over thinking it? 

image

@brandonreid ah yes thanks for the accessibility reminder. I believe this should be pretty similar to PTabs. I'll look at that once the other comments are resolved.

src/components/Accordion/PAccordion.vue
znicholasbrown: I buy that - thanks for the explanation (and also for pioneering the SFC generics support for this lib)

znicholasbrown:
Hmm I see what you mean @pleek91 - for the header differentiation, perhaps the cue is another horizontal rule between heading and content by default? Then a slotted caller could remove it (as in your screenshot above) but the base component would remain visually viable. 

For the entire header being the button, I don't see those as completely incompatible but would require slotted callers to stop event propagation if there are interactive elements in the header itself. I'll leave that decision up to @brandonreid but I appreciate the explanation!

brandonreid:
We discussed in slack. Let's opt to make the whole accordion header a clickable button and adopt a design principle of no other nested interactive elements within.  We'll modify the dashboard design.

src/components/Accordion/PAccordion.vue
pleek91: While updating this I realized I could simplify the generics I had and remove the need for the string casting. I had a more complex generic to try and support `defineSlots` but ultimately abandoned the slot types because of the complexity it added and also it just wouldn't compile.

src/components/Accordion/PAccordion.vue
pleek91: Updated this to use a v-model. By default no sections are open.

znicholasbrown (approved):

brandonreid (approved):

PR #825, opened 2023-06-22
RFC: PSelectOption add the ability to target or omit specific values

pleek91:
# Description
Adds two actions to PSelectOption. One for selecting only the highlighted value (removes any other values) and one for selecting all values except the highlighted value. 

This loom goes into a little more detail about how this works and also explains how most of this is just a launching point. 
https://www.loom.com/embed/8f8c6648b7f54e2bbb5ee99c147f9bd9

image
image

### DIFF
--- a/demo/sections/components/Select.vue
+++ b/demo/sections/components/Select.vue
@@ -67,6 +67,18 @@
     {
       label: 'Three', value: '3',
     },
+    {
+      label: 'Four', value: '4',
+    },
+    {
+      label: 'Five', value: '5',
+    },
+    {
+      label: 'Six', value: '6',
+    },
+    {
+      label: 'Seven', value: '7',
+    },
   ]
 
   const exampleOptionsGrouped = [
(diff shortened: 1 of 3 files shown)


znicholasbrown (commented):
Generally I like this UX but can we find a design that doesn't nest buttons? And perhaps some more fitting icons, though I'll admit I haven't looked to see if there are any that make sense

pleek91:
> Generally I like this UX but can we find a design that doesn't nest buttons? And perhaps some more fitting icons, though I'll admit I haven't looked to see if there are any that make sense

From an html perspective there are no nested buttons. But I'm guessing you mean how the click target for the option itself (which is a div) surrounds the actions I've added here. Broadly agree though that is going to be difficult to get around. 

The icons are not finalized. Should have included more in the PR description. Will add a loom explaining these changes.

znicholasbrown (approved):

pleek91:
Closing this as the ux issues were never resolved.

PR #832, opened 2023-07-06
Enhancement: Better date formatting of selected date and time in p-date-input

pleek91:
# Description
Specifically set out to solve the issue where the selected time was not shown unless you opened the input. Made some overall changes to how we format and display the text/date/time. Feedback welcome

No value before
image

No value after
image

No value and `showTime` before
image

No value and `showTime` after
image

Date selected before
image

Date selected after
image

Date and time selected before
image

Date and time selected after
image

### DIFF
--- a/src/components/DateInput/PDateButton.vue
+++ b/src/components/DateInput/PDateButton.vue
@@ -33,16 +33,26 @@
 
   const props = defineProps<{
     date: Date | null | undefined,
+    showTime?: boolean,
   }>()
 
   const buttonElement = ref<HTMLButtonElement>()
   const el = computed(() => buttonElement.value)
-  const displayValue = computed(() => {
-    if (props.date) {
-      return format(props.date, 'MM/dd/yyyy')
+
+  const dateTimeFormat = computed(() => {
+    if (props.showTime) {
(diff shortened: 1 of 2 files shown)


brandonreid (approved):
❤️

znicholasbrown (approved):
Approving because this is strictly better - one question but not blocking. Thanks @pleek91

src/components/DateInput/PDateButton.vue
```diff
@@ -33,16 +33,26 @@
...
   const buttonElement = ref<HTMLButtonElement>()
   const el = computed(() => buttonElement.value)
+
+  const dateTimeFormat = computed(() => {
+    if (props.showTime) {
+      return 'MMM do, yyyy \'at\' h:mm a'
+    }
+
+    return 'MMM do, yyyy'
+  })
```
znicholasbrown, line 48: Any chance of localization? This is strictly better than the current version but leans in to US style dates and I'm assuming `date-fns` isn't taking that into account when passing a specific format

pleek91: Correct it does not. I briefly looked into that but decided that was more of a change than I wanted to make today. This aligns it with the rest of our products (which also do not consider localization).

PR #833, opened 2023-07-07
BugFix: PDateRange use correct times

pleek91:
# Description
Fixes an issue where selecting the same start and end date would result in a date difference of 0 minutes. Should also fix issues where ranges of dates wouldn't include some data when using those dates directly for filters and queries. 

Closes https://github.com/PrefectHQ/prefect/issues/7917

This change forces the start date to always be 00:00:00 and the end date to be 59:59:59 by using date-fns startOfDate and endOfDay utilities.

Side quest: Also updated the formatting when displaying the dates to match PDateInput

### DIFF
--- a/src/components/DateInput/PDateButton.vue
+++ b/src/components/DateInput/PDateButton.vue
@@ -52,7 +52,7 @@
       return format(props.date, dateTimeFormat.value)
     }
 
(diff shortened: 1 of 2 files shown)


brandonreid (approved):
Lgtm!

PR #856, opened 2023-07-24
Feature: Error boundary component

znicholasbrown:
I am once again asking for your support of a `PErrorBoundary` component


pleek91:
@znicholasbrown I feel like last time we tried this it didn't actually work like we expected. But maybe things have improved since then? 

Either way, I'm open to this. But curious what the use case is driving this. I get how it works but why specifically would we use it.

znicholasbrown:
The specific use-case I had in mind is to wrap 3rd party components (things like Stripe forms) whose errors we don't have great control over when mounting. Instead of relying on the 3rd party to have appropriate error handling (and accepting that it might not be particularly help to a user), we can define our own AND we can throw a standardized error message for alerting purposes

cc @zhen0 since we discussed this in the context of logrocket alerting i believe

brandonreid (approved):

znicholasbrown:
Sorry I meant to comment on this and instead took it out of draft - I want to work with @zhen0 to make sure this actually fits the use case so this'll likely change a bit

zhen0 (approved):
Thanks @znicholasbrown

PR #859, opened 2023-07-25
add root background colors

brandonreid:
This adds the root background colors. Some bg colors, for things like inputs and buttons, will have their own semantic tokens in fallowing PR's, at which point I'll remove the previous custom bg utilities in favor of the new ones.

### DIFF
--- a/demo/App.vue
+++ b/demo/App.vue
@@ -34,8 +34,6 @@
 
 .app { @apply
   text-foreground
-  bg-background-900
-  dark:bg-background-50
   grid
   max-h-screen
 }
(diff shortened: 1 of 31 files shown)


pleek91:
@brandonreid this PR needs to be against the `redesign` branch and not `main`

brandonreid:
> @brandonreid this PR needs to be against the `redesign` branch and not `main`

Good looking out @pleek91 , fixed

pleek91:
@brandonreid as part of this PR I'd expect to see the existing background colors removed (both background and probably foreground).

brandonreid:
> @brandonreid as part of this PR I'd expect to see the existing background colors removed (both background and probably foreground).

Yeah, in the PR description I address that. There are background colors for things like inputs, states, and interactive elements that will have their own tokens. I can do all backgrounds at once, will just be a larger PR? Or I figured I'll address removing those once the component tokens have made it through.

pleek91 (approved):

PR #860, opened 2023-07-26
Add semantic tokens

brandonreid:
Adding the known designed set of root level semantic tokens. Component level tokens that are needed will be added as said component is updated in future PR's.


src/styles/index.css
```diff
@@ -139,6 +139,29 @@
...
+    --p-color-event-dense: #A564F9;
+
+    --p-color-sentiment-positive: #22C55E;
+    --p-color-sentiment-neutral: #70707B;
+    --p-color-sentiment-negative: #D92D21;
+
+    /* Text */
+    --p-color-text-default: #FFFFFF;
+    --p-color-text-subdued: #ADADAD;
+    --p-color-text-inverse: var(--p-color-bg-0);
```
pleek91, line 161: If we changed the color of bg-0 would we want this text to match it? If not I'd just put the hex value here.

pleek91 (approved):
Approving as to not block but two comments

src/styles/index.css
brandonreid: I set the inverse to the lite mode text color and vise versa, rather than the background color. 👍

PR #861, opened 2023-07-26
Base input and accordion

brandonreid:
Updated styles utilizing tokens for the base input and accordion components

### DIFF
--- a/src/styles/index.css
+++ b/src/styles/index.css
@@ -175,9 +175,9 @@
     --p-color-input-border-invalid: var(--p-color-sentiment-negative);
     --p-color-input-text-invalid-icon: var(--p-color-sentiment-negative);
 
-    --p-color-input-activated-bg: var(--p-color-active);
-    --p-color-input-activated-border: var(--p-color-active);
-    --p-color-input-activated-text: var(--p-color-text-default);
+    --p-color-input-checked-bg: var(--p-color-active);
+    --p-color-input-checked-border: var(--p-color-active);
+    --p-color-input-checked-text: var(--p-color-text-default);
 
     /* Legacy Tokens – To be removed before releasing version 2.0 */
     --foreground: var(--200);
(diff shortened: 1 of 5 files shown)


znicholasbrown (commented):
Is `hue` the right word for this token? Reading through these tokens it looks like we're saying:
color-focus-color. Perhaps something like focus-highlight?

src/styles/index.css
```diff
@@ -161,6 +165,20 @@
...
+    --p-color-input-text: var(--p-color-text-default);
+    --p-color-input-placeholder: var(--p-color-text-subdued);
+
+    --p-color-input-bg-invalid: var(--p-color-input-bg);
+    --p-color-input-border-invalid: var(--p-color-sentiment-negative);
+    --p-color-input-text-invalid-icon: var(--p-color-sentiment-negative);
+
+    --p-color-input-activated-bg: var(--p-color-active);
+    --p-color-input-activated-border: var(--p-color-active);
+    --p-color-input-activated-text: var(--p-color-text-default);
```
brandonreid, line 180: The activated input tokens are for checked checkbox/radios. Accidentally added those, can remove if we wanna.

znicholasbrown: Can we swap that for `active` then? I believe that's the CSS semantic term instead of `activated`

brandonreid: I think you're thinking of `checked`, which I can swap for that here.. I can't imagine another input type that would have a sort of activated state, so..

brandonreid: Updated!

znicholasbrown: Ah yeah you're right, sorry. Yea I'm mostly trying to make sure we can map tokens onto their semantic meanings for consistency. We'll be grateful for that later I think

znicholasbrown (approved):

src/styles/index.css
brandonreid: Totally! More than welcomed!

PR #862, opened 2023-07-27
Buttons and breadcrumbs

brandonreid:
This PR essentially refactors buttons with new design tokens and reduces the options down to what we use, and intend to continue using, in the app.

Focus rings have also been updated as applying the previous approach lead to primary buttons having a hugging focus ring that essentially just made the button look a little bigger. This makes focus rings consistent. For buttons I was able to make it so they only show when focusing with keyboard. The way the inputs are built make it difficult since there is no `focus-visible-within` pseudo class and my attempts to make it work failed.

Then breadcrumbs were hit up real quick.


src/components/Button/PButton.vue
```diff
@@ -24,32 +27,31 @@
...
+    },
     to: {
       type: [String, Object] as PropType<RouteLocationRaw>,
       default: undefined,
     },
+    small: Boolean,
+    activated: Boolean,
+    disabled: Boolean,
+    loading: Boolean,
   })
```
pleek91, line 55: While we're updating components we should convert props over to how we do it in the other repos. 
```typescript
  const props = defineProps<{
    kind: ButtonKind,
    icon?: Icon,
    iconAppend?: Icon,
    to?: RouteLocationRaw,
    small?: boolean,
    activated?: boolean,
    disabled?: boolean,
    loading?: boolean,
  }>()
```

src/components/Button/PButton.vue
```diff
@@ -24,32 +27,31 @@
...
+    iconAppend: {
+      type: String as PropType<Icon>,
+      default: undefined,
+    },
     to: {
       type: [String, Object] as PropType<RouteLocationRaw>,
       default: undefined,
     },
+    small: Boolean,
+    activated: Boolean,
```
pleek91, line 52: What is `activated`?

src/components/Button/PButton.vue
```diff
@@ -9,9 +9,12 @@
   >
     <div class="p-button__content">
       <template v-if="icon">
-        <PIcon :icon="icon" class="p-button__icon" />
+        <PIcon :size="small ? 'small' : 'large'" :icon="icon" class="p-button__icon" />
```
pleek91, line 12: nit: Since you're using this twice I'd abstract this into a computed

```
const iconSize = computed(() => props.small ? 'small' : 'large')
```

src/types/button.ts
```diff
@@ -0,0 +1 @@
+export type ButtonKind = 'default' | 'primary' | 'danger' | 'primary--danger' | 'flat'
```
pleek91, line 1: nit: I'd keep to a single dash for `primary--danger`

```suggestion
export type ButtonKind = 'default' | 'primary' | 'danger' | 'primary-danger' | 'flat'
```

Also, can you give me an example where we would need "danger" vs "primary-danger"? It would be nice to not have two "danger" styles. But mostly having one two word kind is annoying me lol.

src/styles/index.css
```diff
@@ -146,8 +147,8 @@
     --p-color-divider: #51525C;
 
     --p-color-active: #1F647A;
-    --p-color-focus: #175CD3;
-    --p-color-focus-halo: rgba(23, 92, 211, 0.4);
+    --p-color-focus-ring: #276EE7;
+    --p-color-focus-ring-offset: var(--p-color-bg-0);
```
pleek91, line 151: hmm this is interesting. I'd expect this to be the bg of the layer its on. Or transparent. Can this be transparent?

src/styles/index.css
```diff
@@ -179,6 +183,62 @@
...
+    /* Button */
+    --p-color-button-default-bg: transparent;
+    --p-color-button-default-border: #70707B;
+    --p-color-button-default-text: var(--p-color-text-default);
+    --p-color-button-default-bg-hover: var(--p-color-button-default-bg);
+    --p-color-button-default-border-hover: var(--p-color-text-subdued);
+    --p-color-button-default-text-hover: var(--p-color-button-default-text);
+    --p-color-button-default-bg-active: #141517;
+    --p-color-button-default-border-active: var(--p-color-button-default-border-hover);
+    --p-color-button-default-text-active: var(--p-color-button-default-text-hover);
```
pleek91, line 195: Seems like most of these are the default tokens. Do we need all these button specific tokens? Possibly these are just here in case we change our mind and want these to have different values?

src/types/size.ts
pleek91, line 1: I feel like this is used for more than just buttons and probably cannot be deleted.

pleek91: At least not yet

src/styles/index.css
brandonreid: I'll try this again, but I believe transparent wasn't working.. doesn't make sense that it wouldn't though..

src/components/Button/PButton.vue
brandonreid: So the `activated` variant is really only for default buttons when they're part of a button group.
image

I could probably just apply this as overrides from the button group though, huh..

src/styles/index.css
pleek91: `bg-0` is probably fine if transparent doesn't work but it stuck out to me as unexpected so thought I'd ask.

src/components/Button/PButton.vue
pleek91: hmm, can this be a separate `kind` rather than its own prop? Then I think this make sense. I'd just make it `active` as the kind.

src/types/button.ts
brandonreid: Yeah, I feel the annoyance. A standard danger button is for any ol' destructive activity. So you might see one of these:
image

Whereas a primary-danger button functions like a primary button, it has prominent visibility as it constitutes the "next step in a journey". So in a confirmation modal, etc. you'd have a primary-danger button.

image

brandonreid: I think `dangerous` can actually be a prop that applies to any button type

pleek91: Ah so just `<p-button dangerous>` and `<p-button kind="primary" dangerous>`?

brandonreid: right right, which also enables `<p-button icon="TrashIcon" kind="flat" dangerous />`, which I'll add styles for

pleek91: At that point I might go back to accepting just the names like `<p-button primary dangerous>`

pleek91: I like that. Either with or without the kind property.

brandonreid: Yeah.. Only combo you wouldn't really want is `flat primary`.. will wrestle with that

pleek91: Agreed. But in that case I think primary just takes over. 

I've seen quite a few button/theme systems designed that way.

src/styles/index.css
brandonreid: Yeah so it's a nested shadow thing, the offset is another layered shadow on top to separate the outer shadow, not like a gradient thing like I was trying to imagine.

pleek91: Ha yeah that makes sense. This is fine the.

src/styles/index.css
brandonreid: Yeah for sure. Again, buttons will be a unique beast in this regard. Overall we should be able to avoid pattern level tokens like this (in this case, pattern variant level). In this case, since we're doing it, this provides us total themability if our default buttons end up having a bg in the future, etc.

src/types/size.ts
brandonreid: There are a couple references to it in prefect-ui-library, but only used as forward props to p-button instances. So in those cases, I'll have to swap that prop to just a `small` boolean.

brandonreid: No other references to it in prefect-design. Not sure why it was named generically like that..

pleek91: Probably had plans for it's use more generally that never happened. Thanks for checking!

src/components/Button/PButton.vue
brandonreid: I removed it for now, will reintroduce it when I hit up button groups.. probably in this same PR but after. Might try and do it from that component anyway.

src/components/Button/PButton.vue
```diff
@@ -88,208 +78,172 @@
...
-    'p-button-md': props.size === 'md',
-    'p-button-lg': props.size === 'lg',
-    'p-button-xl': props.size === 'xl',
+    'p-button--icon-only': props.icon && !slots.default,
+    'p-button--primary': props.primary,
+    'p-button--danger': props.dangerous,
+    'p-button--flat': props.flat,
+    'p-button--small': props.small,
+    'p-button--icon-prepend': props.icon && slots.default,
+    'p-button--icon-append': props.iconAppend && slots.default,
```
pleek91, line 87: These icon classes that check the slot. Whats the thought there?

brandonreid: So the i guess it just keeps less classes applying. I'm realizing now, there's no harm in `icon-only` and either of these classes being applied at the same time since they all do the same thing (reduce left/right padding in some form). Feels a little cleaner to keep them? Or is dumping the extra checks and `useSlots` method call not worth?

pleek91: ah I see. I'm not sure `slots.default` will be reactive like you think. But this is fine for now. We can address if it becomes an issue.

src/components/Button/PButton.vue
```diff
@@ -88,208 +78,172 @@
...
 }
-
-.p-button-md { @apply
-  text-base
-  px-4
-  py-2
+.p-button--primary.p-button--danger:not(:disabled):active {
+  background-color: var(--p-color-button-primary-danger-bg-active);
+  border-color: var(--p-color-button-primary-danger-border-active);
+  color: var(--p-color-button-primary-danger-text-active);
```
pleek91, line 196: We might want to come back through and make these BEM compatible at some point. But should be fine for now. Some of these selectors have a high specificity.

brandonreid:
Opting to open this up for review and just get it merged. Won't worry about the build on every PR -> `rebrand` until we're getting closer to release, just to keep the PR's smaller.

PR #863, opened 2023-07-28
propigate updated buttons throughout repo

brandonreid:
- Updates the `p-button-group` component.
- Updates button usage throughout the repo to use the updated props.
- A couple minor tweeks to demo stuff is sprinkled in here as well.

### DIFF
--- a/demo/components/AppNavigationBar.vue
+++ b/demo/components/AppNavigationBar.vue
@@ -1,5 +1,5 @@
 <template>
-  <p-navigation-bar class="app-navigation-bar p-background" :class="classes.root" v-bind="{ layout }">
+  <p-navigation-bar class="app-navigation-bar" :class="classes.root" v-bind="{ layout }">
     <template #prepend>
       <slot name="prepend" v-bind="{ layout }">
         <template v-if="layout == 'horizontal'">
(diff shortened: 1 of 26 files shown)


demo/components/AppNavigationBar.vue
```diff
@@ -1,9 +1,9 @@
 <template>
-  <p-navigation-bar class="app-navigation-bar" :class="classes.root" v-bind="{ layout }">
+  <p-navigation-bar class="app-navigation-bar p-background" :class="classes.root" v-bind="{ layout }">
```
znicholasbrown, line 2: Try to keep only BEM classes in template elements, even demo ones; that gives us very clear places to maintain styles

src/components/Button/PButton.vue
```diff
@@ -235,6 +235,11 @@
   py-1
 }
 
+.p-button--small .p-button__icon { @apply
+  max-w-[20px]
+  max-h-[20px]
```
znicholasbrown, line 240: I think we should be using relative units instead of px here

src/components/ButtonGroup/PButtonGroup.vue
```diff
@@ -3,9 +3,9 @@
     <template v-for="button in options" :key="button.label">
       <PButton
         :disabled="button.disabled"
-        :size="size"
-        :inset="button.value !== modelValue"
+        :small="small"
         class="p-button-group__button"
+        :class="{ 'p-button-group__button--selected': button.value === modelValue }"
```
znicholasbrown, line 8: We should avoid computed objects in the template cause it's tough to extend/maintain. I'd put this into a computed variable in the script instead

znicholasbrown (commented):
a few comments, overall lgtm

PR #866, opened 2023-07-28
style code components

brandonreid:
Style the code syntax compnents

### DIFF
--- a/src/components/CodeInput/PCodeInput.vue
+++ b/src/components/CodeInput/PCodeInput.vue
@@ -132,18 +132,19 @@
 }
 
 .p-code-input,
-.p-code-input__view,
-.p-code-input__control,
-.p-code-input__textarea,
-.p-code-input__line-numbers-wrapper {
+.p-code-input .p-code-input__view,
+.p-code-input .p-code-input__control,
+.p-code-input .p-code-input__textarea,
+.p-code-input .p-code-input__line-numbers-wrapper {
   font-size: inherit;
   font-family: inherit;
   line-height: inherit;
 }
 
-.p-code-input__control  {
+.p-code-input .p-code-input__control  {
   scrollbar-gutter: stable;
   scrollbar-width: thin;
+  border-radius: 0px;
 }
 
 .p-code-input__textarea,
@@ -154,7 +155,8 @@
 }
 
 .p-code-input { @apply
-  bg-background-500
+  bg-code
+  text-code
   font-mono
   overflow-hidden
   p-0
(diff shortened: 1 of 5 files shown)


src/components/CodeInput/PCodeInput.vue
```diff
@@ -132,18 +132,19 @@
...
 
 .p-code-input,
-.p-code-input__view,
-.p-code-input__control,
-.p-code-input__textarea,
-.p-code-input__line-numbers-wrapper {
+.p-code-input .p-code-input__view,
+.p-code-input .p-code-input__control,
+.p-code-input .p-code-input__textarea,
+.p-code-input .p-code-input__line-numbers-wrapper {
```
znicholasbrown, line 138: Why the increase in specificity here? We shouldn't need to do this for BEM classes

src/components/CodeInput/PCodeInput.vue
```diff
@@ -132,18 +132,19 @@
...
   font-size: inherit;
   font-family: inherit;
   line-height: inherit;
 }
 
-.p-code-input__control  {
+.p-code-input .p-code-input__control  {
```
znicholasbrown, line 144: Same question here

src/styles/index.css
```diff
@@ -523,6 +535,7 @@ html {
 html,
 body {
   background-color: var(--p-color-bg-0);
+  color: var(--p-color-default)
```
znicholasbrown, line 538:
```suggestion
  color: var(--p-color-text-default)
```
I don't see a variable for `p-color-default`, should this be using the default text color?

src/components/CodeInput/PCodeInput.vue
brandonreid: So here we're adding styles for other components, the flat specificity appeared to not be overriding the style declarations in the components. Why was this build this way? There's a lot of small questions like this that I'll continue to encounter, haha. I will take another look at this though.

src/styles/index.css
brandonreid: Ah super good catch, ty

src/components/CodeInput/PCodeInput.vue
znicholasbrown: Hm I'm not sure what you mean by other components, most of these are declared in the template but let's pair on this so I can explain how it works if you get stuck, this is one of the more complicated components cause it's doing some interesting stuff (like the invisible input + overlay)

brandonreid: Hey my comment there was clearly burnt-out-Friday-evening. I’ll take a closer look at this for sure, will reach out if I’m confused. It probably just needs to be moved down the file rather than increased selectors.

znicholasbrown (approved):

PR #871, opened 2023-08-01
Fix: Allow editing of long strings in the PCodeInput component

zhen0:
We've had reports of a bug in our code input.  The video before shows the issue - I can't click into a long string input:
https://github.com/PrefectHQ/prefect-design/assets/40272060/d4c4a2a8-79ac-4abf-af23-ced7ac1a2369

This PR refactors how we set the width of the textarea in the PCodeInput to set it according to the width of the input/the width of the sibling view-container. 

https://github.com/PrefectHQ/prefect-design/assets/40272060/fdea2b01-73d2-4b69-ab17-7863a051a23f


src/components/CodeInput/PCodeInput.vue
```diff
@@ -111,6 +112,7 @@
     return {
       textarea: {
         height: `${lineHeight.value * lines.value}px`,
+        width: `${internalValue.value.length * 8 + 10}px`,
```
zhen0, line 115: This calculation is a bit ad hoc.  I'd hoped to use 'ch' but that caused my UI to crash so px was the best alternative.

src/components/CodeInput/PCodeInput.vue
```diff
@@ -105,12 +110,14 @@
     textArea: {
       'p-code-input__textarea--show-line-numbers': props.showLineNumbers,
     },
+
```
pleek91, line 113:
```suggestion
```

pleek91 (approved):

PR #876, opened 2023-08-04
style date pickers

brandonreid:
Styles the DateInput, DatePicker, and DateRangeInput components. I also hit up some buttons I missed previously as my find-all was for `<p-button` but there were a number of `<PButton` instances out there.

### DIFF
--- a/src/components/Button/PButton.vue
+++ b/src/components/Button/PButton.vue
@@ -276,9 +276,9 @@
 .p-button--selected,
 .p-button--selected:not(:disabled):hover,
 .p-button--selected:not(:disabled):active {
-  background-color: var(--p-color-button-activated-bg);
-  border-color: var(--p-color-button-activated-border);
-  color: var(--p-color-button-activated-text);
+  background-color: var(--p-color-button-selected-bg);
+  border-color: var(--p-color-button-selected-border);
+  color: var(--p-color-button-selected-text);
   cursor: default;
 }
 </style>
--- a/src/components/Button/PButton.vue
+++ b/src/components/Button/PButton.vue
@@ -33,6 +33,7 @@
   const props = defineProps<{
     primary?: boolean,
     flat?: boolean,
+    selected?: boolean,
     icon?: Icon,
     iconAppend?: Icon,
     dangerous?: boolean,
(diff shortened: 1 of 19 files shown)


src/components/Button/PButton.vue
```diff
@@ -33,6 +33,7 @@
   const props = defineProps<{
     primary?: boolean,
     flat?: boolean,
+    selected?: boolean,
```
znicholasbrown, line 36: What's `selected` in relation to a button? i believe this was moved from the button group component but I don't think it makes sense here. There's no model value on this component so all this does is provide alternate styling. I'd move this back to the button group.

brandonreid: A number of additional components in the date picker also utilize a selected style, so at that point I opted to centralize these styles in the button. In the context of a small module like this, it felt reasonably to allow for selected style buttons: 

https://github.com/PrefectHQ/prefect-design/assets/6776415/9c235cde-80ea-4bd0-a420-843242d81437

So if keeping it, I should surface a model value for selected so it can be used like `<p-button v-model:selected="condition" />`, huh?

znicholasbrown: I don't think you need to surface a v-model for it since the button doesn't change that value (so there's never a need to emit the value back to the parent) - `selected` isn't a typical button state. Perhaps `active` as a boolean attribute is more closely in live with how you intend to use this and would give you the opportunity to merge the logic here with what already exists

brandonreid: Hmm. So I had this prop named `active`, but changed it to `selected` just to be in alignment with the token name. The token name is selected > active since active is used as a token modifier for active states on things.

The experience of the button in this state does consistently constitute a selected button option in a group of others, just in the case of the date picker, there's lots of options spread out in a different layout than the ButtonGroup component.

Thoughts?

znicholasbrown: Hm ok I think this is sort of muddying the waters of responsibility for this component but short of changing how you intend to use it I think this is fine. I'd prefer that button groups (which aren't semantic elements afaik) would instead be styled radio buttons or tablist elements and wouldn't actually be wrapping the PButton component. Since `selected` doesn't add any accessibility attributes it's probably ok

znicholasbrown (approved):

src/components/Button/PButton.vue
brandonreid: That's a great point on button groups actually.. I agree on the sentiment as well. To sleep well tonight I made a [ticket](https://github.com/PrefectHQ/prefect-design/issues/886) for addressing bucket groups. Will push on for now.

PR #879, opened 2023-08-04
style markdown renderer and tables

brandonreid:
As I updated styles for the markdown renderer, I notices tables got some special styles for a sticky header. I had previously removed the bg from tables with the intent to make them sit in cards and let the card handle bg-colors, but seeing this and auditing how we use tables a bit, I added the bg's back to the table on the header and body independently so they work with the sticky header.

### DIFF
--- a/src/styles/index.css
+++ b/src/styles/index.css
@@ -545,12 +545,6 @@ body {
   color: var(--p-color-text-default)
 }
 
-* {
-  --duration: 50ms;
-  --timing: ease;
-  transition: background-color var(--duration) var(--timing), color var(--duration) var(--timing);
-}
-
 *::selection {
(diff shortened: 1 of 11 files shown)


src/styles/index.css
```diff
@@ -545,12 +545,6 @@ body {
   color: var(--p-color-text-default)
 }
 
-* {
```
brandonreid, line 548: @znicholasbrown I assume this wildcard transition was added to animate the light/dark modes. Lmk if you feel strongly about it, but I felt it was a bit overreaching for an affect that is pretty noticeable and thus probably doesn't warrant the transition.

znicholasbrown: It's fine to remove

PR #888, opened 2023-08-07
Radio inputs, tabs, tags, wizard

brandonreid:
Updated styles for radio inputs, tabs, tags, and wizard.

### DIFF
--- a/src/components/Tag/PTag.vue
+++ b/src/components/Tag/PTag.vue
@@ -47,7 +47,7 @@
 
 <style>
 .p-tag { @apply
-  gap-2
+  gap-1
   inline-flex
   items-center
   rounded-md
@@ -59,8 +59,7 @@
   color: var(--p-color-tag-text);
 }
 
-.p-tag__icon,
-.p-tag__icon svg { @apply
+.p-tag__icon { @apply
   w-4
   h-4
 }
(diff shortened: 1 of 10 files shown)


src/components/Tag/PTag.vue
```diff
@@ -37,29 +37,39 @@
...
-  text-sm
+  text-sm;
+  background-color: var(--p-color-tag-bg);
+  border-color: var(--p-color-tag-border);
+  color: var(--p-color-tag-text);
 }
 
-.p-tag__icon { @apply
+.p-tag__icon,
+.p-tag__icon svg { @apply
```
pleek91, line 63: Seems like this `svg` selector shouldn't be needed if `p-icon` is working properly. Might be worth looking into.

pleek91 (approved):

src/components/Tag/PTag.vue
brandonreid: Yeah that's what I had figured too but didn't dig in, wasn't working properly in the tags demo. I'll go ahead and scope it out before merging this.

brandonreid: So `p-icon` didn't have any sort of max width/height. I went ahead and added those in for that component. We'll just want to keep an eye out for any icons that get squeezed in the UI after the update.

brandonreid: Ah you know what though.. Maybe `p-icon` shouldn't have a max width/height like that but enforce it's standard sizes.. `p-tag` is setting a width and height when it should just be setting the icon to `size="small"`. I'll update that real quick in the rebrand branch directly.

PR #890, opened 2023-08-08
style toggles

brandonreid:


src/components/Toggle/PToggle.vue
```diff
@@ -41,6 +41,7 @@
 
   const props = defineProps<{
     modelValue: boolean | null | undefined,
+    neutralStates?: boolean,
```
pleek91, line 44: I would drop `states` here and just name the prop `neutral`. Since there is already a `state` prop I'm worried these will feel like they are related when they are not.

pleek91 (approved):

PR #892, opened 2023-08-08
PTerminal adding output for command line.

robfreedy:
This adds an "output" prop to the PTerminal component to display command line outputs underneath the "commands" property. 

It also adds a $ to each command in p terminal to. The copy functionality does not include the $ in this case.

### DIFF
--- a/src/components/Terminal/PTerminal.vue
+++ b/src/components/Terminal/PTerminal.vue
@@ -10,8 +10,13 @@
     <div class="p-terminal__code">
       <template v-for="(line, index) in commands" :key="index">
         <p class="p-terminal__code-line">
-          {{ line }}
+          $ {{ line }}
           <span class="p-terminal__cursor" />
+        </p>
+      </template>
+      <template v-for="(line, index) in outputs" :key="index">
+        <p class="p-terminal__code-line">
+          {{ line }}
         </p>
       </template>
     </div>
@@ -27,13 +32,21 @@
 
   const props = defineProps<{
     command: string | string[],
+    output: string | string[],
   }>()
 
   const commands = computed(() => asArray(props.command))
+  const outputs = computed(() => asArray(props.output))
 
   async function copy(): Promise<void> {
     const copyText = commands.value.join(' && ')
     await navigator.clipboard.writeText(copyText)
+
+    showToast('Copied!', 'success')
+  }
+
+  async function copyLine(index: string): Promise<void> {
+    await navigator.clipboard.writeText(commands.value[index])
 
     showToast('Copied!', 'success')
   }
--- a/src/components/Terminal/PTerminal.vue
+++ b/src/components/Terminal/PTerminal.vue
@@ -32,7 +32,7 @@
 
   const props = defineProps<{
     command: string | string[],
-    output: string | string[],
+    output?: string | string[] | null | undefined,
   }>()
 
   const commands = computed(() => asArray(props.command))
--- a/src/components/Terminal/PTerminal.vue
+++ b/src/components/Terminal/PTerminal.vue
@@ -44,12 +44,6 @@
 
     showToast('Copied!', 'success')
   }
-
-  async function copyLine(index: string): Promise<void> {
-    await navigator.clipboard.writeText(commands.value[index])
-
-    showToast('Copied!', 'success')
-  }
 </script>
 
 <style>
--- a/demo/sections/components/Terminal.vue
+++ b/demo/sections/components/Terminal.vue
@@ -7,7 +7,7 @@
     </template>
 
     <template #terminal>
-      <p-terminal command="perfect config set PREFECT_API_URL" />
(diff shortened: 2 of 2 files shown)


pleek91:
@robfreedy whats the motivation for this? The code looks fine. But this component is designed to show commands and allow the user to copy them. I'm concerned adding the output will confuse users since there is one copy button and it might be confusing what is being copied.

robfreedy:
@pleek91 This change was to (at least initially) help with some of the onboarding work that we are working on (see notion here: https://www.notion.so/prefect/Onboarding-Takeover-Design-Doc-cc49e96e7eed4070b32baed603abfda3). 

The idea was that when a user logs in on CLI, creates a work pool/deployment, etc. we could have an output message saying something like "Waiting for deployment" or "Deployment Successfully created" as a validation before the user moves on to the next step. 

Figured this could be used in other areas as well but the output message should be optional so hoping it wouldn't change any of the existing terminals in the UI.

robfreedy:
@pleek91 @brandonreid Open to use p-message to handle the dynamic messaging. Do y'all still want to add the dollar sign to the beginning of the terminal line? On copy this would leave out the dollar sign.

brandonreid:
@robfreedy && @pleek91 

I don't have an opinion against the $ if that's what users expect.

Maybe something like this would work (I have these changes local)?
image
 I think the idea of in-context feedback is pretty cool, but definitely agree that we want to be clear about what is static vs. responsive to the user's actions. Even this, I'm not sure if it communicates "this is the outcome of what you just did", like maybe the "output" label isn't accurate here. Maybe the example in the demo file is confusing me, like is that text what would be used?

pleek91:
@brandonreid I think some interactivity could be nice. But I'm also confused at the uses case. In this particular case it sounds like we're instructing/messaging the user that we're waiting on them to run the command. Which doesn't seem like output from the terminal command we're asking them to run.

znicholasbrown:
> @robfreedy I'm not sure using the terminal is the best place for dynamic user messaging. These are generally static pieces of documentation. Could something like `p-message` be used instead? @brandonreid any thoughts here?

@pleek91 I don't think this is a rule, even if it's historically how we've used this component internally. `PTerminal` at least hints that the component has the look and feel of a command line, which is made of inputs and outputs, and I don't think there's anything that says this should be documentation. 

I motivated this change to give us a little more flexibility:
1. Allow prefixing any commands to simulate a real environment ($ for bash, % for mac terminal etc) - this PR doesn't quite do this yet
2. Allow hinting at the output from a command or series of commands; I think there are a few ways we could handle this but I like @robfreedy's version as a first pass. Ideally the input/output props would be index-tied, something like:

```
const in = ['ls', 'cd docs', 'printenv']
const out = [['docs', 'src'], null, 'FOO=bar']
```

src/components/Terminal/PTerminal.vue
```diff
@@ -10,10 +10,15 @@
     <div class="p-terminal__code">
       <template v-for="(line, index) in commands" :key="index">
         <p class="p-terminal__code-line">
-          {{ line }}
+          $ {{ line }}
```
znicholasbrown, line 13: Instead of a hard-coded $, I'd prefer we went with a prefix string property, defaulting to empty:

```suggestion
          <span v-if="prefix">{{ prefix }} </span> {{ line }}
          
          ...
          
          
          const props = defineProps<{
            ...
            prefix?: string  
          }>()
```

pleek91: We probably want all terminal components to have the same prefix across all our use cases though right? I feel like having this be a prop (or at least not having a default) will mean this will get copy/paste all over and likely to be missed when not pasting. 

Also, what's an example of a different prefix that might be used?

znicholasbrown: Maybe - as you mentioned above, sometimes we use these for more display purposes and other times we use it to spell out a command or series of commands a user should run. The former probably doesn't require a prefix and the latter probably does. We're pretty bash-heavy in examples but having some flexibility here leaves the door open for any sort of prompt. Using the example from above might look something like:

```
$ ls
docs src
$ cd docs
/docs $ printenv
...
```

or the same thing in powershell:

```
PS C:\> ls
docs src
PS C:\> cd C:\docs
PS C:\docs> Get-ChildItem env:
...
```

znicholasbrown:
Hm, one other option would be to wrap one of the code components (PCode or PCodeHighlight) in PWindow to give the same effect but then I'm really not sure what the purpose of this component is - all it does is wrap a string in PWindow with a copy button... for the record I like the design @brandonreid shared but it is a more opinionated change design-wise and doesn't convey "terminal" in quite the same way I envision for this

src/components/Terminal/PTerminal.vue
znicholasbrown: writing that powershell example was actually painful, maybe we shouldn't do this after all so we don't encourage powershell usage

PR #895, opened 2023-08-11
add design token docs

brandonreid:


demo/sections/DesignTokens.vue
```diff
@@ -0,0 +1,400 @@
...
+            <strong>Level 1:</strong> Add the class <p-code inline>
+              p-background
+            </p-code> directly to a container to apply the level 1 background color. Used on cards by default.
+          </p>
+          <div class="p-background mt-4 p-6 rounded-md">
+            <p>
+              <strong>Level 2:</strong> Adding the class <p-code inline>
+                p-background
+              </p-code> directly to a container will automatically apply the level 2 background color when it's nested within a level 1 background, using a simple <p-code inline>
+                .p-background + .p-background {}
```
pleek91, line 233: I don't think this is correct. It should just be `.p-background .p-background` right?

demo/sections/index.ts
```diff
@@ -4,6 +4,7 @@ import { layouts } from '@/demo/sections/layouts'
 
 export const sections: Section = {
   components,
+  designTokens: () => import('./DesignTokens.vue'),
```
pleek91, line 7: Seems like the dynamic menu stuff isn't expecting camel case stuff here. Ugly but I think this will fix it
image

```suggestion
 'Design Tokens': () => import('./DesignTokens.vue'),
```

PR #906, opened 2023-08-14
Feature: PDialog component

pleek91:
# Description
Adds a new `PDialog` component which is a reactive wrapper around the [dialog](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) element. It has a single default slot which binds a close method. This component does not add any opinionated styling and is designed to be styled at implementation. 

## Props
- open - `boolean` (required)
- autoClose - `boolean`
- modal - `boolean`

### DIFF
--- a/src/components/Dialog/PDialog.vue
+++ b/src/components/Dialog/PDialog.vue
@@ -9,7 +9,9 @@
 </template>
 
 <script lang="ts" setup>
+  import { useGlobalEventListener } from '@prefecthq/vue-compositions'
   import { onMounted, onUnmounted, ref, watch } from 'vue'
+  import { isHtmlElement } from '@/utilities/html'
 
   const props = defineProps<{
     open: boolean,
@@ -34,14 +36,15 @@
     }
 
     dialog.value.show()
+    setTimeout(() => addOnClickListener())
   }
 
   function close(): void {
-    if (!dialog.value) {
-      return
+    if (dialog.value) {
+      dialog.value.close()
     }
 
-    dialog.value.close()
+    removeOnClickListener()
   }
 
   function syncOpenValue(): void {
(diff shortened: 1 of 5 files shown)


src/components/Dialog/PDialog.vue
```diff
@@ -0,0 +1,81 @@
...
+  function syncOpenValue(): void {
+    if (!dialog.value) {
+      return
+    }
+
+    emit('update:open', dialog.value.open)
+  }
+
+  function onClick(event: MouseEvent): void {
+    if (!props.autoClose) {
```
stackoverfloweth, line 56: I would have expected `auto-close` to trigger close on any outside click? Isn't that how it works on a-modal and a-popover?

stackoverfloweth (approved):
@pleek91 love this

src/components/Dialog/PDialog.vue
pleek91: I think that's how this is working. If you add the `auto-close` prop if you click anywhere on the backdrop (technically anywhere outside of the `p-dialog__content` div) it'll call the `close` method.

src/components/Dialog/PDialog.vue
```diff
@@ -0,0 +1,81 @@
...
+    if (open) {
+      show()
+      return
+    }
+
+    close()
+  })
+
+  onMounted(() => {
+    dialog.value?.addEventListener('close', syncOpenValue)
```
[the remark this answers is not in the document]
pleek91: I had the same thought but the only one that exists is for global (document.addEventListener) stuff. I might go make one for the current component calls (useEventListener).

brandonreid (approved):

src/components/Dialog/PDialog.vue
stackoverfloweth: interesting. I'm surprised that the click handler on `dialog` gets triggered by anything on the page.

I also noticed that your check for clicks inside the dialog only checks if the target === the dialog element. Wouldn't it be better to check if `dialog.value.contains(event.target)`?

pleek91: hmm so when a dialog is opened as a modal (using `showModal`) there is a `::backdrop` that covers the entire page. So anywhere you can click you trigger a click on the dialog element. So that's why I'm checking that the click is the dialog and not one of its children. (checking dialog.contains would mean clicking on the content of the dialog would close it as well). BUT that doesn't work if its not a modal, so I'll need to account for that. Good catch!

pleek91: @stackoverfloweth fixed!

PR #910, opened 2023-08-16
move components to a plugin

brandonreid:
I was having issues where the custom component classes weren't available in the demo files. Importing this lib into the vue-charts repo, I also had issues where the build wasn't recognizing the custom component classes.

Looking in this repo, I saw that the `forms` plugin was imported twice, and presume the forms weren't being imported from out tailwindPlugins file, but were directly. 

This change fixes my issues, feels a little cleaner as well.


pleek91:
@brandonreid I'd prefer if we could combine the theme and utilities plugins into a single plugin. We're unlikely to ever install one and not the other so its a simpler install with a single plugin file.

brandonreid:
> @brandonreid I'd prefer if we could combine the theme and utilities plugins into a single plugin. We're unlikely to ever install one and not the other so its a simpler install with a single plugin file.

Man I was struggling to make this happen this morning, but it clicked now. Combined into one file.

public/tailwindPlugin.js
```diff
@@ -1,7 +1,6 @@
 /* eslint-disable @typescript-eslint/explicit-function-return-type */
 /* eslint-disable @typescript-eslint/no-require-imports */
 /* eslint-disable @typescript-eslint/no-var-requires */
-const forms = require('@tailwindcss/forms')
```
pleek91, line 4: I'm not quite understanding why we're removing the forms plugin from the prefect-design plugin. Wont this have negative effects assuming we're depending on the forms plugin always being installed whenever the prefect-design plugin is used?

pleek91: I thought forms was added to the config? That's what I'm seeing locally which appears to be valid looking at the types.

brandonreid: It is, yeah. Forms was previously added twice. Once in the tailwindPlugin file, where it was the third argument and I don't think it was actually doing anything, then once in the tailwind.config file, where it was actually being imported and subsequently bundled with the build.

pleek91: Okay, trust you're correct 👍

PR #919, opened 2023-08-21
style sticky header when stuck

brandonreid:
Requires https://github.com/PrefectHQ/vue-compositions/pull/294


src/types/reactivity.ts
```diff
@@ -0,0 +1 @@
+export type Getter<T> = () => T
```
znicholasbrown, line 1: Are we not at the appropriate Vue version for this export?

znicholasbrown (approved):

src/types/reactivity.ts
brandonreid: I'm not sure Getter is actually defined in the latest vue. I see [MaybeRefOrGetter](https://vuejs.org/api/utility-types.html#maybereforgetter) of course, but don't see Getter.

znicholasbrown: Oh weird ok

Issue #922, opened 2023-08-24
Bug: Overflow menu gradient obfuscates content and looks bad


znicholasbrown:
![Screenshot 2023-08-24 at 2 32 27 PM](https://github.com/PrefectHQ/prefect-design/assets/27291717/c3935a0b-d0de-4750-95da-ef7f3592be73)

![Screenshot 2023-08-24 at 3 21 38 PM](https://github.com/PrefectHQ/prefect-design/assets/27291717/fccd362e-e8be-4ad8-bd07-20683bc4ba30)

![Screenshot 2023-08-24 at 3 21 31 PM](https://github.com/PrefectHQ/prefect-design/assets/27291717/b778804b-592c-4321-a791-c8c1fa948202)

Gradient should only apply if container is scrollable

znicholasbrown:
Resolved by #923

PR #923, opened 2023-08-24
fix: menu gradient in content

diegolikescode:
Hello! Take a look, I've fixed the #922 problem.

The problem is that it renders 2x, in the first (I believe it renders the input first) ``endVisible.value`` is always false, so when you add the first element and beyond it remains false. This is why the second time the content loads, since it has ``options``, it loads the way it should.

I hope it was helpful! Maybe a better solution is to render only once, but this is a solution for the problem raised in the issue


zhen0:
Thanks for the contribution.  This looks fixed in the deploy preview to me but I'll let @znicholasbrown or @pleek91 have final approval.

znicholasbrown (approved):
This is great, thank you @diegolikescode !

PR #924, opened 2023-08-25
Enhancement: Allow PContextNavItem to be a button by making `to` optional

pleek91:
# Description
Makes the `to` prop optional. This follows how PButton works so that the nav item can either be a router link or a button.


znicholasbrown (approved):
I'm surprised we don't need any additional styling for the base HTML button, nice

PR #935, opened 2023-08-30
Resolve v2 feedback

brandonreid:
Apologies for the larger PR, I did try and break the commits up as I went though!

### DIFF
--- a/src/components/ButtonGroup/PButtonGroup.vue
+++ b/src/components/ButtonGroup/PButtonGroup.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="p-button-group">
-    <template v-for="button in options" :key="button.label">
+    <template v-for="button in options" :key="button.value">
       <p-button
         :disabled="button.disabled"
         :small="small"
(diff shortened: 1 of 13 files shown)


src/layouts/PLayoutTable/PLayoutTable.vue
```diff
@@ -35,23 +35,22 @@
...
+  const { rootMargin, boundingElement } = toRefs(props)
   const options: Getter<UsePositionStickyObserverOptions> = () => {
-    if (!boundingElement?.value) {
-      return {}
-    }
-
     return {
-      boundingElement: boundingElement.value,
+      rootMargin: rootMargin?.value ?? undefined,
+      boundingElement: boundingElement?.value ?? undefined,
```
pleek91, line 50: Won't these be undefined anyway? Doesn't seem like the coalesce is necessary

src/components/ButtonGroup/PButtonGroup.vue
```diff
@@ -9,7 +9,9 @@
         :icon="button.icon"
         @click="select(button.value)"
       >
-        {{ button.label }}
+        <template v-if="button.label" #default>
+          {{ button.label }}
+        </template>
```
pleek91, line 14: If you're making label optional you'll need to expose this slot

pleek91 (changes requested):
The button group slot needs to be exposed otherwise everything else LGTM

pleek91 (approved):

PR #952, opened 2023-09-08
Bugfix: Don't flex default slot

znicholasbrown:
There's not really a reason to flex the slot, downstream consumers should be responsible for that


brandonreid (approved):

pleek91 (approved):
This seems likely to cause issues with implementations that depend on this flex existing. Might need to update a few places.

znicholasbrown:
I thought the same but there are only 2 components that rely on this behavior and neither are currently being used 🤷🏻

Issue #960, opened 2023-09-11
Bug: `PSelectOptions`/`PSelect` don't expose virtual scroller events


znicholasbrown:
This means it's difficult to understand when to fetch more options for API-driven options that exceed the current window since it's not possible to know when the final item is in view without a special observer

PR #962, opened 2023-09-12
Enhancement: Bubble up the bottom event from the virtual scroller in PSelect & PCombobox

pleek91:
# Description
Since there's already an infinite scroll component in PSelectOptions, bubbling that up through the parent components means implementations can utilize that event.


znicholasbrown:
Closes https://github.com/PrefectHQ/prefect-design/issues/960

znicholasbrown (approved):

PR #984, opened 2023-10-10
Bugfix: `PSelect` empty message

znicholasbrown:
`PSelect` was making an implicit assumption that the model value for the component would exist in the list of options and thereby render the `empty-message` slot if a value was set but not present. This is problematic in circumstances where options are fetched on-demand from an async source or when paginating the options list.


pleek91 (approved):
So simple, yet so subtle

PR #985, opened 2023-10-11
Fix stepper disabled

stackoverfloweth:
p-stepper does not currently accept `disabled` prop. It sort of works by passing that prop through to the underlying p-number-input component but the buttons used in prepend and append are stick clickable and still appear to modify the value. 

this change makes `disabled` an expected prop, uses to disable stepper buttons and still passes through to underlying p-number-input


pleek91 (approved):
Nice, thank you!

PR #987, opened 2023-10-12
[PEmptyState] Don't render slots(and their styles) when slots are not used.

collincchoy:
## Description
The `<p-empty-state>` component comes with implicit margins in most (all but the icon) slots even if a consumer doesn't want to use that slot. This leads to things like [this](https://github.com/PrefectHQ/nebula-ui/blob/6ba913394f150f127789c2fdc24eff41b57b2ef0/src/components/SearchDialogEmptyState.vue#L26-L30) when someone notices the extraneous space and wants to rid themself of it. 

```
.search-dialog-empty-state .p-empty-state__heading,
.search-dialog-empty-state .p-empty-state__actions,
.search-dialog-empty-state .p-empty-state__corner-label { @apply
  hidden
}
```

## What it does
Don't render the slot containers (which have the overbearing styles) if no slot contents are provided for those respective slots. This makes using each slot optionally more lenient for a consumer.

With the following code:
image

| **slot** | **implicit margins** |
| --- | --- |
| corner-label | image |
| heading | image |
| description | image |
| actions | image |

### DIFF
--- a/src/components/EmptyState/PEmptyState.vue
+++ b/src/components/EmptyState/PEmptyState.vue
@@ -1,26 +1,26 @@
 <template>
   <p-card class="p-empty-state">
-    <div class="p-empty-state__corner-label">
+    <div v-if="$slots.cornerLabel" class="p-empty-state__corner-label">
       <slot name="corner-label" />
     </div>
     <div
       class="p-empty-state__text"
     >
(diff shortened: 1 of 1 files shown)


pleek91 (approved):
Nice, this stuff always sneaks through. Thanks for fixing!

PR #989, opened 2023-10-13
update slot props on PTabSelect to be consistent with PTabNavigation

### LINKED ISSUE #988: p-tabs heading slot has in consistent props
the p-tabs component switches from `PTabNavigation` to `PTabSelect` on media.sm

the goal of these components was to produce a consistent dx regardless of screen size so the developer wouldn't have to worry about providing slot content to both permutations. However, there is an issue with the heading slot. In `PTabNavigation`, the "-heading" slot looks like this

```html
<slot
  :name="`${kebabCase(tab.label)}-heading`"
  v-bind="{ tab, index }"
>
```

the "same" slot in `PTabSelect` looks like this

```html
<slot
  :name="`${kebabCase(option.label)}-heading`"
  v-bind="{ tab: option.label, index }"
>
```

Notice how the value for `tab` goes from `{ label: string }` to just `string`. This means if the developer provides the following slot content

```html
<template #heading="{ tab }">
  {{ tab?.label }} ❤️
</template>
```

That will work fine on bigger screens, but on smaller screens `tab` is a string, so `tab?.string` will be `undefined` 💔

In fact this is already testable on the demo page

1.  go to https://prefect-design.netlify.app/components/tab
2. shrink your viewport width to media.sm or smaller
3. notice how only the icon shows, and the text "Tab One" disappears

Screenshot 2023-10-12 at 8 26 37 PM

The ideal solution to this would be a simple update to `PTabSelect` to bind `tab` to the entire `option`, which more closely mirrors the tab from `PTabNavigation`

image

However, that would be a breaking change for any component that expected `tab` to be a string.

stackoverfloweth:
this change could theoretically introduce a breaking change if any thing was using PTabSelect and expected `tab` to be string. Theoretically any slot that accounted for this would have a case for `tab` being `{ label: string }`, which will always be the case now and shouldn't cause any issues.

closes #988

### DIFF
--- a/src/components/Tabs/PTabSelect.vue
+++ b/src/components/Tabs/PTabSelect.vue
@@ -10,9 +10,9 @@
     <template #option="{ option, index }">
       <slot
         :name="`${kebabCase(option.label)}-heading`"
-        v-bind="{ tab: option.label, index }"
+        v-bind="{ tab: option, index }"
       >
-        <slot name="heading" v-bind="{ tab: option.label, index }">
+        <slot name="heading" v-bind="{ tab: option, index }">
           {{ option.label }}
         </slot>
       </slot>


pleek91 (approved):
Yeah this makes sense to me. Looking at past PRs seems like this just go missed

PR #1003, opened 2023-10-31
Enhancement: Improved sanitization

znicholasbrown:
This PR updates the markdown renderer and sanitize html components to better improve their security posture.

`PSanitizeHtml` no longer allows `form` elements and explicitly forbids `script`, `style`, and `iframe` elements instead of relying on the default behavior of DOMPurify.

It also explicitly strips the following HTML attributes: 
- `action`
- `method`
- `onclick`
- `onmouseover`
- `onload`
- `data`

`PMarkdownRenderer`'s parser now also adds `rel="noopener"` to all markdown-created links (ones generated by `[link](https://some-link)`) as an added layer of protection against tab/window hijacking

### DIFF
--- a/src/components/SanitizeHtml/PSanitizeHtml.vue
+++ b/src/components/SanitizeHtml/PSanitizeHtml.vue
@@ -13,13 +13,15 @@
     html: string,
   }>()
 
-  const forbiddenAttrs = ['style']
+  const forbiddenAttrs = ['style', 'action', 'method', 'onclick', 'onmouseover', 'onload', 'data']
+  const forbiddenTags = ['script', 'iframe', 'form']
   const useProfiles = { svg: true, html: true }
 
   const sanitize = (text: string): string => {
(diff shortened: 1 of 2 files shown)


src/components/SanitizeHtml/PSanitizeHtml.vue
```diff
@@ -13,13 +13,15 @@
     html: string,
   }>()
 
-  const forbiddenAttrs = ['style']
+  const forbiddenAttrs = ['style', 'action', 'method', 'onclick', 'onmouseover', 'onload', 'data']
```
brandonreid, line 16: Any concerns about `onmousedown` or `onmouseup`? Or maybe those are covered by click.

znicholasbrown: Done

brandonreid (approved):

PR #1012, opened 2023-11-08
Enhancement: Allow passing sanitization config to `PSanitizeHtml` component

znicholasbrown:
This will provide downstream consumers a way to override the component defaults as necessary without needing to re-implement the component.

Note: I've also updated allowed types for both `PSanitizeHtml` and `PUnwrap` due to the change in overload coming from the config.

### DIFF
--- a/src/components/SanitizeHtml/PSanitizeHtml.vue
+++ b/src/components/SanitizeHtml/PSanitizeHtml.vue
@@ -10,17 +10,18 @@
   import { PUnwrap } from '@/components/Unwrap'
 
   const props = defineProps<{
-    html: string,
+    html: string | HTMLElement | DocumentFragment,
+    config?: dompurify.Config,
   }>()
 
   const forbiddenAttrs = ['style', 'action', 'method', 'onclick', 'onmouseover', 'onload', 'data', 'onmousedown', 'onmouseup']
   const forbiddenTags = ['script', 'style', 'iframe', 'form']
   const useProfiles = { svg: true, html: true }
 
-  const sanitize = (text: string): string => {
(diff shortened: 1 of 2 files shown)


pleek91 (approved):

src/components/SanitizeHtml/PSanitizeHtml.vue
```diff
@@ -10,17 +10,18 @@
...
   const forbiddenAttrs = ['style', 'action', 'method', 'onclick', 'onmouseover', 'onload', 'data', 'onmousedown', 'onmouseup']
   const forbiddenTags = ['script', 'style', 'iframe', 'form']
   const useProfiles = { svg: true, html: true }
 
-  const sanitize = (text: string): string => {
+  const sanitize = (text: string | HTMLElement | DocumentFragment): string | HTMLElement | DocumentFragment => {
     return dompurify.sanitize(
       text,
-      {
+      props.config ?? {
```
pleek91, line 24: nit: this would be nice rather than declaring the default inline here. Could probably combine with the consts above and just declare a single const with the correct type. 
```
props.config ?? defaultConfig
```

znicholasbrown: Ah shoot I merged before this comment came through, sorry!

pleek91: No that's my bad, I accidentally submitted the review without this comment and had to go back and add it.

znicholasbrown: https://github.com/PrefectHQ/prefect-design/pull/1013

PR #1017, opened 2023-11-15
add sort placement in table layout header

brandonreid:
Creates a dedicated space in the table header for sort options, so it can be separated with a divider from other filter options:

image

### DIFF
--- a/demo/sections/layouts/LayoutTable.vue
+++ b/demo/sections/layouts/LayoutTable.vue
@@ -12,16 +12,17 @@
           </template>
 
           <template #header-start>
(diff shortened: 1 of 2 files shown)


znicholasbrown (approved):
is this a new pattern you're hoping to introduce? i've not seen us do this yet but i might have missed it

brandonreid:
> is this a new pattern you're hoping to introduce? i've not seen us do this yet but i might have missed it

Yeah a design pattern at least. I'll make a PR to update the implementations in the app as well.

PR #1019, opened 2023-11-16
Add incident severity icons

collincchoy:
Adds 6 new Prefect icons. 

[Figma](https://www.figma.com/file/FEX3VwXS3FdJnS8b75VXHi/Incidents?type=design&node-id=50-7523&mode=design&t=gWg1PnygnMih1PqI-4)

image


pleek91 (approved):

brandonreid (approved):
Looks good! I don't believe you need the `fill="none"` on the `<svg>` elements? nbd of course, g2g

collincchoy:
> Looks good! I don't believe you need the `fill="none"` on the `<svg>` elements? nbd of course, g2g

Hm yeah iirc setting fill on the `<svg>` elements would cascade down to children so maybe it'd be more accurate to set fill="currentColor" there and then only need to use fill within children that _need_ to override? 🤔 

altho looks like a lot of other svg's here set currentColor directly on children and leave fill="none" on the `<svg>` so feels like if we wanna prefer that then we could address all at once.

PR #1028, opened 2023-12-01
Update picker for date inputs

pleek91:
# Description
Updates the existing PDateInput and PDateRange input components to have the updated date picker experience

## Date Input
| Description | Before | After |
|--------|--------|--------|
| Empty Picker | image | image |
| Date selected | image | image | 

## Date Range Input
| Description | Before | After |
|--------|--------|--------|
| Empty Picker | image | image |
| Range selected | image | image | 

## Clearing values
Both the date and date range inputs support clearing the current value if by clicking the X button within the input itself.
image


src/components/Calendar/PCalendar.vue
pleek91, line 1: This component used to just be the days of the month. And then in other places we were adding the buttons and the ability to change the year/month. 

I've rolled all that functionality into these calendar components so that when using `PCalendar` you get the complete ux.

src/utilities/position.ts
```diff
@@ -118,4 +138,8 @@ export const rightTop: PositionMethod = function(target: DOMRect, content: DOMRe
     top,
     left,
   }
-}
\ No newline at end of file
+}
+
+// export const rightMiddle: PositionMethod = function(target, content, container) {
+//   const top = target.top - target.height / 2
+// }
```
znicholasbrown, line 145:
```suggestion
```

src/components/DateRangeInput/PDateRangeInput.vue
```diff
@@ -1,133 +1,93 @@
...
-    get() {
-      return props.startDate ?? null
-    },
-    set(value) {
-      emit('update:startDate', value)
-    },
-  })
+  // this dummy ref is bound to PDatInput's v-model but isn't directly used for anything
+  // this component tracks its own dates
+  const dummy = ref()
```
znicholasbrown, line 77: This looked like a dev relic till I saw this comment - maybe rename this var to `internalModelValue` to avoid confusion? then the comment becomes obsolete potentially

src/components/DateRangeInput/PDateRangeInput.vue
```diff
@@ -1,133 +1,93 @@
...
 
-  const internalEndDate = computed({
-    get() {
-      return props.endDate ?? null
-    },
-    set(value) {
-      emit('update:endDate', value)
-    },
-  })
+  const dateFormat = 'MMM do, yyyy'
```
znicholasbrown, line 79: Perhaps this format should be localized

znicholasbrown (approved):
no concerns here, a couple of comments otherwise lgtm

src/components/DateRangeInput/PDateRangeInput.vue
pleek91: I'm not apposed to that. Its just a little different. An internalModelValue I would expect to match the modelValue prop which there is none in this component (there's a startDate and endDate v-model). Also this literally isn't used for anything. This component literally overrides all the functionality of the PDateInput via its slots. But PDateInput requires a v-model. So this is only here to satisfy that constraint. IMO `internalModelValue` doesn't convey that. But will update if you feel strongly.

src/components/DateRangeInput/PDateRangeInput.vue
pleek91: That's probably a good idea. Though at this time I'm not wanting to change how that works as that would effect a number of components.

PR #1034, opened 2023-12-06
Chore: Update PSelectOptions to handle its own selection logic

pleek91:
# Description
Small refactor so that PSelectOptions can be used on its own without reproducing all of the logic for highlighting and selecting values.

### DIFF
--- a/src/components/Select/PSelect.vue
+++ b/src/components/Select/PSelect.vue
@@ -17,7 +17,6 @@
           :class="classes.control"
           :disabled="disabled"
           v-bind="attrs"
-          :options="selectOptions"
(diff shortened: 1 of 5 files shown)


src/components/Select/PSelect.vue
```diff
@@ -134,6 +130,10 @@
 
   const modelValue = computed({
     get() {
+      if (multiple.value) {
+        return props.modelValue ?? []
+      }
+
```
pleek91, line 136: small bugfix. Noticed this component wasn't working correctly with `modelValue: null | undefined` and `multiple: true`

znicholasbrown (approved):

PR #1037, opened 2023-12-07
Add a slot to p-divider for a centered item with space around.

collincchoy:
image

### DIFF
--- a/demo/sections/components/Divider.vue
+++ b/demo/sections/components/Divider.vue
@@ -1,9 +1,13 @@
 <template>
-  <ComponentPage title="Divider" :demos="[{ title: 'Divider' }]">
+  <ComponentPage title="Divider" :demos="[{ title: 'Divider' }, { title: 'With Slot' }]">
     <template #divider>
       <span>above</span>
       <p-divider />
       <span>below</span>
+    </template>
+
+    <template #with-slot>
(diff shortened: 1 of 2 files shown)


brandonreid (approved):
Without doing any work to find out, I wonder if there are any accessibility considerations to be had around two `<hr>` elements divided by something? I think it's a good call to include this in the divider pattern though, since we actually do this in a number of places.

collincchoy:
> Without doing any work to find out, I wonder if there are any accessibility considerations to be had around two `<hr>` elements divided by something? I think it's a good call to include this in the divider pattern though, since we actually do this in a number of places.

it's a good thought. my prior implementation was more "1 line (from a border)" + "some content" with the 2 positioned on top of one another. which i've seen (anecdotally) as common in other implementations. also saw 1 impl w/ an aria-hidden on the line element but... I'm not really sure why that'd be helpful to a sr.. 

Worth noting I altered to this approach because the other one relies on overlaying with a non-transparent background behind the slotted content. Setting that background out-of-the-box becomes the tricky part tho and forcing the consumer to deal with that seems not great.

def curious if there's some weirdness w/ this some-where/how.

pleek91:
Merging this since I'm about to do a release. Think we all agree on the implementation.

PR #1038, opened 2023-12-07
Feature: Add a composition for using theme tokens

pleek91:
# Description
Enables reactive access to theme tokens like
```
const tokens = useThemeTokens()

const styles = computed(() => ({
  color: tokens.colorTextDefault
})
```


src/compositions/useThemeTokens.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  watch(theme, () => Object.assign(tokens, getTokens()))
+
+  return tokens
+}
+
+// this makes more sense to just infer the return type
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+function getTokens() {
+  const style = getComputedStyle(document.documentElement)
```
znicholasbrown, line 17: `getComputedStyle` and all these `getPropertyValue` calls might be non-trivial if lots of components were to use the composition. It might make sense to instantiate the reactive var/watcher outside the composition context so repeated calls to the composition share the same reactive variable.

znicholasbrown (approved):
One suggestion, not really blocking and potentially a premature optimization

znicholasbrown (approved):

src/compositions/useColorTheme.ts
```diff
@@ -1,5 +1,6 @@
 import { useLocalStorage } from '@prefecthq/vue-compositions'
-import { onMounted, onUnmounted, ref, onBeforeMount, Ref } from 'vue'
+import { ref, Ref } from 'vue'
+import { globalExists } from '..'
```
znicholasbrown, line 3: Mind cleaning up this import?

znicholasbrown (approved):

PR #1039, opened 2023-12-08
Fix useThemeTokens returning empty tokens on first render

pleek91:
# Description
Fixes the issue where tokens would be empty strings when the page first loads.

### DIFF
--- a/src/compositions/useThemeTokens.ts
+++ b/src/compositions/useThemeTokens.ts
@@ -1,14 +1,24 @@
 import { reactive, watch, readonly } from 'vue'
 import { useColorTheme } from '@/compositions/useColorTheme'
 
-function factory(): () => Readonly<ReturnType<typeof getTokens>> {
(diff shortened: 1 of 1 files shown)


src/compositions/useThemeTokens.ts
```diff
@@ -1,14 +1,24 @@
...
+
+  setTokens()
 
-  watch(theme, () => Object.assign(tokens, getTokens()))
+  watch(theme, () => setTokens())
+
+  function setTokens(): void {
+    requestAnimationFrame(() => {
+      Object.assign(tokens, getTokens())
+    })
```
znicholasbrown, line 18: What's the thinking behind the animation frame here? I'd expect something more on the lines of:
```ts
document.addEventListener("DOMContentLoaded", () => {
   Object.assign(tokens, getTokens())
})
```

as a better proxy for loaded stylesheets but maybe I'm misunderstanding the problem

znicholasbrown (approved):

PR #1048, opened 2023-12-11
Feature: PDateRangeSelect

pleek91:
# Description
The date range select is a new component for selecting date ranges either of a fixed date/time or a relative time span like "Past 15 minutes". 

It has a single v-model that returns a value based on the user's selection. If a relative value is selected the number of seconds in the selected span is returned. If a fixed range is selected the start and end dates are returned. This value can be type guarded with the `modelValue.type` which is either `span` or `range`

Recommend checking out the [deploy preview ](https://deploy-preview-1048--prefect-design.netlify.app/components/date-range-select)and playing around this the component. 

Relative span
image
image

Fixed date range
image
image

### DIFF
--- a/demo/sections/components/DateRangeSelect.vue
+++ b/demo/sections/components/DateRangeSelect.vue
@@ -1,0 +1,44 @@
+<template>
+  <ComponentPage title="Date Range Select" :demos="[{ title: 'Date Range Select' }]">
+    <template #description>
+      <DemoState v-model:state="exampleState" v-model:disabled="disabled" />
+    </template>
+
+    <template #date-range-select>
+      <p-content>
+        <p-date-range-select v-model="value" />
+        <p-code inline>
+          value: {{ JSON.stringify(value) }}
+        </p-code>
+        <div class="flex justify-start gap-4">
+          <p-checkbox v-model="isClearableDate" :disabled="disabled" label="Clearable" />
+        </div>
+        <div class="flex gap-4 w-1/2">
+          <p-label label="Min Value" :message="minDate ? format(minDate, 'MMM do, yyyy h:mm a') : ''">
+            <p-native-date-input v-model="minDate" :disabled="disabled" :state="exampleState" />
+          </p-label>
+
+          <p-label label="Max Value" :message="maxDate ? format(maxDate, 'MMM do, yyyy h:mm a') : ''">
+            <p-native-date-input v-model="maxDate" :disabled="disabled" :state="exampleState" />
+          </p-label>
+        </div>
+      </p-content>
+    </template>
+  </ComponentPage>
+</template>
+
+<script lang="ts" setup>
+  import { State } from '@/types'
+  import { format } from 'date-fns'
+  import { ref } from 'vue'
+  import ComponentPage from '@/demo/components/ComponentPage.vue'
+  import DemoState from '@/demo/components/DemoState.vue'
+
+  const exampleState = ref<State>()
+  const disabled = ref(false)
+
+  const value = ref()
+  const minDate = ref<Date | null>(null)
+  const maxDate = ref<Date | null>(null)
+  const isClearableDate = ref(false)
+</script>
--- a/demo/sections/components/DateRangeSelect.vue
+++ b/demo/sections/components/DateRangeSelect.vue
@@ -6,12 +6,12 @@
 
     <template #date-range-select>
       <p-content>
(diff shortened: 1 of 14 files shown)


src/components/DateInput/PDateInput.vue
pleek91, line 1: Updated because the PDatePicker component has the floating styles now. I think "Picker" implies its not designed to be put directly on the page.

demo/sections/components/DateRangeSelect.vue
```diff
@@ -0,0 +1,44 @@
...
+    </template>
+
+    <template #date-range-select>
+      <p-content>
+        <p-date-range-select v-model="value" v-bind="{ min, max }" />
+        <p-code inline>
+          value: {{ JSON.stringify(value) }}
+        </p-code>
+        <div class="flex justify-start gap-4">
+          <p-checkbox v-model="isClearableDate" :disabled="disabled" label="Clearable" />
```
zhen0, line 14: Does this do anything?

src/components/DateRangeSelect/PDateRangeSelectOptions.vue
```diff
@@ -0,0 +1,87 @@
...
+  })
+
+  const input = ref<InstanceType<typeof PTextInput>>()
+  const search = ref('')
+
+  const options = computed<SelectOption[]>(() => {
+    const parsed = parseInt(search.value || '1')
+    const unit = isNaN(parsed) ? 1 : parsed
+
+    const spans = [
```
zhen0, line 51: I think it makes sense to include week here too.

demo/sections/components/DateRangeSelect.vue
```diff
@@ -0,0 +1,44 @@
+<template>
+  <ComponentPage title="Date Range Select" :demos="[{ title: 'Date Range Select' }]">
+    <template #description>
+      <DemoState v-model:state="exampleState" v-model:disabled="disabled" />
```
zhen0, line 4: Not sure disabled work here either.  That would be a useful prop to have.

zhen0:
Being able to select the relative time windows and to scroll through them is very cool!

demo/sections/components/DateRangeSelect.vue
pleek91: You're right, another missing prop. Will fix

src/components/DateRangeSelect/PDateRangeSelectOptions.vue
pleek91: I added weeks and also increased the default threshold for removing an option to 2 months. I just want to avoid things like `100 weeks` showing up.

pleek91:
@zhen0 I think I address all of your comments if you'd like to look at this again.

zhen0:
Is it possible to have it so I can select a specific end time?  It keeps resetting when I try to change the end time

https://github.com/PrefectHQ/prefect-design/assets/40272060/9c8cc6c5-dbbe-454e-a8ec-59e8c7af4a4c

zhen0:
Sorry @pleek91 one other thing I didn't catch in my first run through.  The rest is looking nice!

znicholasbrown (approved):

pleek91:
@zhen0 what browser? I don't have any issues changing the times in chrome.

znicholasbrown:
Ah please ignore my review, that was supposed to be a comment; I've seen the behavior @zhen0 is describing. It's also easy to break the input by accidentally pressing > 4 numbers while the year is selected in the calendar view. The only way to escape that that I've found is to refresh

pleek91:
@znicholasbrown I can reproduce the > 4 numbers in the year position. Will investigate. Still unable to reproduce the other issue. Maybe one of you can pair with me on that

znicholasbrown:
@pleek91 select a date range (any date will do) then select the end time hour and press "0" to replicate that one I believe, lmk if that doesn't do it for you

pleek91:
@znicholasbrown ah yep the "0" works. TY

znicholasbrown:
Sorry, I did not mean to close this with my comment, that shouldn't be an option for a maintainer lol

pleek91:
@znicholasbrown and @zhen0 I believe I've fixed the two issues you ran into.

zhen0:
Yep!  LGTM!

znicholasbrown (approved):

PR #1051, opened 2023-12-13
BugFix: Emit the bottom event if the native select is mounted

pleek91:
# Description
If a select or combobox is using the bottom event to async load options no options will appear if a native select is used. This can happen if the `media.hover` check is false. 

Emitting the `bottom` event when the native select is mounted ensures at least the first page of results is shown. 

Note: This is a stop gap to get options showing at all. To restore full functionality to mobile devices we'll need to rethink our mobile strategy for selects

### DIFF
--- a/src/components/Select/PSelect.vue
+++ b/src/components/Select/PSelect.vue
@@ -62,6 +62,7 @@
           :disabled="disabled"
           v-bind="attrs"
           :options="options"
+          @vue:mounted="emit('bottom')"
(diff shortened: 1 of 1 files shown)


zhen0 (approved):

znicholasbrown (approved):
discussed on slack - this will only show the first set of select options on mobile since the event will only ever trigger once but that's ok for now

PR #1052, opened 2023-12-13
Fix PSelect empty message not showing up

pleek91:
# Description
The `multiple` check got changed accidentally to look at the internal model rather than the prop and that caused empty arrays to not be considered as an empty value.


pleek91:
Closes https://github.com/PrefectHQ/nebula-ui/issues/3911

brandonreid (approved):

znicholasbrown (approved):

PR #1055, opened 2023-12-14
Fix: tags input not taking advantage of available width

brandonreid:
The tags input required a fixed width to really take advantage of the `+n` overflow display. This made it so the input was unusually long on the dashboard or never read out the selected tags:

image

This PR makes it so that the tags list has an inline option and will allow the input to scale with the content.

https://github.com/PrefectHQ/prefect-design/assets/6776415/06893d8f-f0ff-409b-a833-5a556cad4c66

https://github.com/PrefectHQ/prefect-design/assets/6776415/0ce71b18-0e10-45cc-91fb-573326f8a387

### DIFF
--- a/demo/sections/components/Tags.vue
+++ b/demo/sections/components/Tags.vue
@@ -60,7 +60,7 @@
         <div>
           <p>Inline</p>
           <div class="border border-default p-2 max-w-full">
-            <p-tag-wrapper :tags="numberArr" justify="left" inline class="border border-red-500" />
+            <p-tag-wrapper :tags="numberArr" justify="left" inline class="border border-default" />
           </div>
         </div>
       </div>
--- a/demo/sections/components/Tags.vue
+++ b/demo/sections/components/Tags.vue
@@ -60,7 +60,7 @@
         <div>
           <p>Inline</p>
           <div class="border border-default p-2 max-w-full">
-            <p-tag-wrapper :tags="numberArr" justify="left" inline class="border border-default" />
+            <p-tag-wrapper :tags="numberArr" justify="left" inline class="border border-red-500" />
           </div>
         </div>
       </div>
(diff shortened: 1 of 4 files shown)


src/components/TagWrapper/PTagWrapper.vue
```diff
@@ -63,44 +64,43 @@
...
-  const calculateOverflow = (): void => {
-    const children = Array.from(
-      container.value?.querySelector('.p-tag-wrapper__tag-container')?.children ?? [])
-      .filter(child => !child.classList.contains('p-tag-wrapper__tag-overflow'),
-      ) as HTMLElement[]
+  const invisibleTagClass = 'p-tag-wrapper__tag--invisible'
+  const hiddenTagClass = 'p-tag-wrapper__tag--hidden'
 
-    const containerBoundingBox = container.value!.getBoundingClientRect()
+  const calculateOverflow = (): void => {
```
brandonreid, line 70: I did a little cleanup in this function but not a full refactor.

pleek91:
Overall code looks fine to me. But when I viewed the demo it seemed like it wasn't working correctly in the tags demo
image

pleek91:
@brandonreid this seems to be working much better. One thing I found while testing in the demo though is if you shrink the inline one down and then expand it it won't always expand to show all the tags. Most reproducible when expanding more slowely. 

image

demo/sections/components/Tags.vue
```diff
@@ -50,11 +51,18 @@
...
+        <p-tag-wrapper justify="right">
           <p-tag v-for="i in 20" :key="i">
             Tag {{ i }}
           </p-tag>
         </p-tag-wrapper>
+
+        <div>
+          <p>Inline</p>
+          <div class="border border-default p-2 max-w-full">
+            <p-tag-wrapper :tags="numberArr" justify="left" inline class="border border-red-500" />
```
pleek91, line 63:
```suggestion
            <p-tag-wrapper :tags="numberArr" justify="left" inline class="border border-default" />
```

pleek91 (approved):
Working good for me!

PR #1056, opened 2023-12-14
Fix: prevent default layout header blowouts

brandonreid:
This is all part of some effort to make the dashboard header more responsive. Prevents flex content from blowing out the container.


pleek91 (approved):
Makes sense to me 👍

PR #1065, opened 2023-12-20
Add sizes for tags

collincchoy:
image

### DIFF
--- a/src/components/Tag/PTag.vue
+++ b/src/components/Tag/PTag.vue
@@ -23,12 +23,13 @@
   import { Icon } from '@/types/icon'
   import { TagValue, normalize } from '@/types/tag'
 
-  const props = defineProps<{
+  const props = withDefaults(defineProps<{
     icon?: Icon,
     label?: string,
     value?: string | TagValue,
     dismissible?: boolean,
-  }>()
+    size?: 'xs' | 'sm',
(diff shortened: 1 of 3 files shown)


pleek91:
@brandonreid

brandonreid (approved):
One comment, otherwise g2g!

src/components/Tag/PTag.vue
```diff
@@ -23,12 +23,13 @@
...
   import { TagValue, normalize } from '@/types/tag'
 
-  const props = defineProps<{
+  const props = withDefaults(defineProps<{
     icon?: Icon,
     label?: string,
     value?: string | TagValue,
     dismissible?: boolean,
-  }>()
+    size?: 'xs' | 'sm',
```
brandonreid, line 31: To be a little more semantically prescriptive, maybe the terms here would be better suited as `'sm' | 'default'`? Or, like buttons, we ended up refactoring to just the default and a bool prop called `small`.

brandonreid: I'm like 98.6% sure we'll never add a `large` for tags

collincchoy: 👍 will shoot for consistency. size -> `small?: boolean`

PR #1066, opened 2023-12-22
Fix: remove height 100% from layouts

brandonreid:
Layouts had a height assignment of 100%, but these weren't neccesary. It made it so that if a layout had content above (e.g. the incidents banner) or below it, the height of the layout would be the height of it's inner+neighboring content, causing too much padding at the bottom of it's inner contents and causing the left navigation to scroll. CSS is weird and hard sometimes.

### DIFF
--- a/src/layouts/PLayoutFull/PLayoutFull.vue
+++ b/src/layouts/PLayoutFull/PLayoutFull.vue
@@ -12,6 +12,7 @@
 .p-layout-full { @apply
   relative
   w-full
+  h-full
 }
 
 .p-layout-full__header { @apply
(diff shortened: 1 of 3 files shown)


src/layouts/PLayoutFull/PLayoutFull.vue
```diff
@@ -12,7 +12,6 @@
 .p-layout-full { @apply
   relative
   w-full
-  h-full
```
pleek91, line 15: Not sure if we even use this one anymore but are you sure about removing this one? This one is for full screen stuff and h-full kinda makes sense here.

pleek91 (commented):
One question but approving assuming you're correct

src/layouts/PLayoutFull/PLayoutFull.vue
brandonreid: I couldn't actually find an example of the full layout in the app, so I just undid this piece to be safe. 👍

pleek91 (approved):

PR #1072, opened 2024-01-02
Add "period" as a date range value type for things like "Today"

pleek91:
# Description
Adds "Today" and as a "period" options to the DateRangeSelect component. Added "Today", "Past hour", 'Past day", and "Past week" as shortcut options visible when first opening the input. 

image


znicholasbrown (approved):
nice

PR #1073, opened 2024-01-03
Enhance types to more easily allow for type guarding date range type

pleek91:
# Description
Enhances the types for DateRangeSelectValue to allow for type safe guarding of the `type` property.


znicholasbrown (approved):
that's some nice type guarding you got there 🤠

PR #1074, opened 2024-01-03
Allow filtering by unit when using the DateRangeSelect's relative time dropdown

pleek91:
# Description
Previously only the number value was considered when typing into the search. Now it will also filter the options based on any string value. So the search `20m` will filter the results to `Past 20 minutes` and `Next 20 minutes`. Also supports `20 min`, `minutes`, etc

Also adds visual indicators of shortcuts for filtering
image
image

### DIFF
--- a/src/components/DateRangeSelect/PDateRangeSelectRelative.vue
+++ b/src/components/DateRangeSelect/PDateRangeSelectRelative.vue
@@ -31,18 +31,20 @@
   const search = ref('')
 
   const options = computed<SelectOption[]>(() => {
-    const parsed = parseInt(search.value || '1')
-    const unit = isNaN(parsed) ? 1 : parsed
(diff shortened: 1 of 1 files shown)


src/components/DateRangeSelect/PDateRangeSelectRelative.vue
```diff
@@ -87,4 +114,15 @@
 .p-date-range-select-relative {
   min-width: 300px;
 }
+
+.p-date-range-select-relative .p-select-option__text { @apply
+  w-full
+}
```
pleek91, line 120: PSelectOption doesn't allow for full width content of its slot. This seemed like the simplest and least risky option for getting the layout necessary without having to refactor PSelectOption and test everything that uses it.

znicholasbrown (approved):

PR #1095, opened 2024-01-23
Feature: `PDraggableList` (WIP)

znicholasbrown:


znicholasbrown:
Not sure how I feel about the name for the base component, might go with `POrderableList` or something similar

src/components/DraggableList/PDraggableList.vue
```diff
@@ -0,0 +1,257 @@
...
+    },
+    item: (index: number) => ({
+      'p-draggable-list__item--over': dragging.value && overIndex.value === index,
+      'p-draggable-list__item--dragging': dragging.value && draggingIndex.value === index,
+    }),
+  }))
+
+  const itemIsDraggable = (index: number): Booleanish => {
+    return draggingIndex.value === index ? 'true' : 'false'
+  }
```
pleek91, line 73: Does vue force the draggable attribute to be shorthand if a boolean is used?

src/components/DraggableList/PDraggableList.vue
```diff
@@ -0,0 +1,257 @@
...
+        @drop="drop"
+        @keydown="handleKeydown($event, i)"
+      >
+        <div class="p-draggable-list__item-handle" @mousedown="handleMouseDown(i)" @mouseup="handleMouseUp">
+          <slot name="handle" v-bind="{ item, index: i }">
+            <PIcon icon="DragHandle" />
+          </slot>
+        </div>
+        <div class="p-draggable-list__item-content">
+          <slot v-bind="{ item, index: i }" />
```
pleek91, line 23: What do you think about binding a delete callback as part of the scope? That way at implementation the logic for deleting doesn't need to be written there as well. Can just grab it from the scope.

pleek91 (commented):
This looks great to me. Actually surprised how simple this turned out. native drag/drop has come a long way.

src/components/DraggableList/PDraggableList.vue
znicholasbrown: Love that - I'll think through if there are other pieces we need to emit from the top level component and/or expose through the scope. At the moment I'm not sure about the level of customization this is giving us for using this on individual fields

src/components/DraggableList/PDraggableList.vue
pleek91: That's weird, like why lol. Anyways carry on!

src/components/DraggableList/PDraggableList.vue
pleek91: I think exposing some of the move item functionality as callbacks would be nice as well. Things like `moveUp`, `moveDown`, `moveToTop`, and `moveToBottom` that way the implementation could add buttons or menu items for that functionality as well.

znicholasbrown:
I think generally this is ok to ship but mobile is still iffy - I'm gonna take another crack at it tomorrow morning and see how far I get; if I'm still stuck I think this is worth merging and opening a ticket to look at mobile instead of letting this hold up downstream feature work

I also wanna pitch a potential set of interactions for multi-select drag and drop, using shift + click or ctrl + click. Curious if that sparks interest in you @pleek91 / @brandonreid - no worries if not, either way I'd like to handle in a follow up

pleek91:
> I think generally this is ok to ship but mobile is still iffy - I'm gonna take another crack at it tomorrow morning and see how far I get; if I'm still stuck I think this is worth merging and opening a ticket to look at mobile instead of letting this hold up downstream feature work
> 
> I also wanna pitch a potential set of interactions for multi-select drag and drop, using shift + click or ctrl + click. Curious if that sparks interest in you @pleek91 / @brandonreid - no worries if not, either way I'd like to handle in a follow up

I think this is good enough to ship and then we can enhance as needed. Good mobile support would be good. And I think if we have multi sort/drag that would be a nice enhancement. But if we can release something that works that we can build off of I could start using this as is.

pleek91 (approved):

PR #1106, opened 2024-02-05
PDraggableList: Only prevent default if taking an action. Add shift to enter and delete

pleek91:
# Description
Makes some tweaks to the keydown handler so that enter and delete are combined with shift. Otherwise if you use the slot and use something like a code input or a textarea adding a newline adds a new item. Also not being able to use the delete is bad. So adding shift to those. I think the arrow keys will also need some tweaking. But leaving those for now.

### DIFF
--- a/src/components/DraggableList/PDraggableList.vue
+++ b/src/components/DraggableList/PDraggableList.vue
@@ -217,14 +217,9 @@
     const altKeyIsPressed = event.altKey
     const shiftKeyIsPressed = event.shiftKey
 
-    const handledKeys = ['Enter', 'Delete', 'ArrowUp', 'ArrowDown']
-    const shouldPreventDefault = handledKeys.includes(event.key)
-
-    if (shouldPreventDefault) {
+    if (shiftKeyIsPressed && event.key === 'Enter') {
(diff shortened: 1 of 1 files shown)


brandonreid:
Yeah the suggested shortcuts when designing were pretty tunnel visioned on text inputs. I wonder if we could have a method for distinguishing? Really only with simple text inputs does the affordance of shortcuts make sense I'm thinking.

For delete, the idea was that if you cleared the input, then pressed delete again, that's when you'd actually delete the added field. I wonder if that should actually be handled at a higher level than here (perhaps all these shortcuts?)

pleek91:
> Yeah the suggested shortcuts when designing were pretty tunnel visioned on text inputs. I wonder if we could have a method for distinguishing? Really only with simple text inputs does the affordance of shortcuts make sense I'm thinking.
> 
> For delete, the idea was that if you cleared the input, then pressed delete again, that's when you'd actually delete the added field. I wonder if that should actually be handled at a higher level than here (perhaps all these shortcuts?)

Yeah I think we should revisit this. But for now this at least makes them usable.

PR #1107, opened 2024-02-07
Enhancement: add lvl-3 bg color token

brandonreid:
Because the schema form UI can reach a third layer of card nesting relatively easily, we're adding a third level bg token that otherwise should be considered undesirable to use. 

Dark mode:
image

Light mode:
image


pleek91 (approved):
Nice, yeah this will help a lot with schemas! TY

PR #1125, opened 2024-02-26
Fix p-divider sometimes has 0 width.

collincchoy:
Fixes small regression from https://github.com/PrefectHQ/prefect-design/pull/1037

|     | **Before** | **After** |
| ---- | ---- | ---- |
| Deployment details well | image | image |
| Artifact page well | image | image |

### DIFF
--- a/src/components/Divider/PDivider.vue
+++ b/src/components/Divider/PDivider.vue
@@ -1,18 +1,25 @@
 <template>
-  <div class="flex items-center gap-x-2">
-    <hr class="p-divider">
+  <div class="p-divider">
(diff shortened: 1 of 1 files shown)


src/components/Divider/PDivider.vue
```diff
@@ -1,18 +1,25 @@
...
     </template>
   </div>
 </template>
 
 <style>
   .p-divider { @apply
+    flex
+    items-center
+    gap-x-2
+    w-full
```
collincchoy, line 19: this `w-full` on the root element is the fix.

PR #1126, opened 2024-02-27
design: add right border to sidebar

aaazzam:
This PR adds a 1px right divider-colored border to the sidebar.

This is one of a series of PRs to better compartmentalize or "bento" the UI. 

This _may_ necessitate removing or shifting the `decorative_iso-pixel-grid_light.svg` background image. These don't conflict outright, but the pixel grid image _feels_ hard to have in a bento world. Would love to find another way to incorporate the brand imagery, but this doesn't feel like a blocking question for now. 

before:
Screenshot 2024-02-26 at 10 46 12 PM

after:
Screenshot 2024-02-26 at 10 46 05 PM

### DIFF
--- a/src/components/NavigationBar/PNavigationBar.vue
+++ b/src/components/NavigationBar/PNavigationBar.vue
@@ -60,6 +60,8 @@
   h-screen
   w-64
   shadow
+  border-r
+  border-divider
 }
 
 .p-navigation-bar--horizontal { @apply


brandonreid:
Thanks for these shots @aaazzam 

In the case of the images you shared there, especially with the main content having the background image that clips, the border totally helps. In the app though, I'm not sure I see it adding to the experience. It consumes white space and creates an unnecessarily hard separation between the nav and content, boxing things in and adding clutter, where without it the separation of concerns is still solid enough and there's more breathing room.

Before:
Screenshot 2024-02-27 at 8 38 54 AM

After:
Screenshot 2024-02-27 at 8 39 43 AM

To get heady with with the bento concept, it actually resists divider lines as an entity in itself. You might imagine a bento box from a sushi restaurant and say "the sushi is divided from the ginger", but if you look closely a more accurate statement would be "the sushi is in a different container than the ginger". Divider lines are sort of unnatural in that sense and contained elements make for stronger visual groupings than divided ones.

image

We've explored bento-ing our UI more as well, like in the design summit prototype:
image
The content header row isn't playing nicely here. True bento design is largely impractical imo, as organizing content cleanly in view like the [slick Apple designs](https://miro.medium.com/v2/resize:fit:1400/format:webp/1*X9b4rlXrkzR-uM9HzxJF8w.jpeg) is challenging and inflexible for an application, but the effort to question dividers in favor of containers still has it's merits.

Totally open to counter arguments and further rational. 🙏

PR #1128, opened 2024-02-29
Standardize gap spacing in p-form to match p-content

pleek91:
# Description
`p-form` has built in spacing for child elements so that forms have consistent spacing. But its set to spacing 6 when we generally use spacing 4. So we have a decent bit of forms that have a `p-content` wrapping all of their content to get the right spacing. Updating `p-form` to use spacing 4 means we can just use the form by itself without a `p-content`.

### DIFF
--- a/demo/sections/components/Form.vue
+++ b/demo/sections/components/Form.vue
@@ -6,17 +6,16 @@
 
     <template #form>
       <p-form @submit="submit">
-        <p-content>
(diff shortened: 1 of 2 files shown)


brandonreid (approved):
It looks like the demo example has a `p-content` container. Took a second to get how that spacing was being applied, but I see now. Might be worth updating the demo? Nbd really.

image

PR #1136, opened 2024-03-07
Draft: LayoutGrid

aaazzam:
Adds a header/toolbar/{leftRail/main/rightRail} layout. 

Note, the left border on the layout is a concession to scope visual design to a single component. Doing this properly IMO would put a border on the sidebar, but skirting that issue for now.

### DIFF
--- a/demo/sections/layouts/LayoutGrid.vue
+++ b/demo/sections/layouts/LayoutGrid.vue
@@ -1,0 +1,57 @@
+<template>
+  <ComponentPage title="Grid Layout" :demos="[{ title: 'Well Layout' }]">
+    <template #well-layout>
+      <p-layout-well-toolbar class="bg-slate-100">
+        <template #header-leading>
+          <div class="h-full">
+            Header Leading
+          </div>
+        </template>
+        <template #header-trailing>
+          <div class="h-full">
+            Header Trailing
+          </div>
+        </template>
+        <template #toolbar-leading>
+          <div class="h-full">
+            Toolbar Leading
+          </div>
+        </template>
+        <template #toolbar-trailing>
+          <div class="h-full">
+            Toolbar Trailing
+          </div>
+        </template>
+        <div class="p-background h-full">
+          <div class="bg-gray-300 layout-well-section__box" />
+          <div class="bg-gray-200 layout-well-section__box" />
+          <div class="bg-gray-300 layout-well-section__box" />
+          <div class="bg-gray-200 layout-well-section__box" />
+          <div class="bg-gray-300 layout-well-section__box" />
+        </div>
+        <template #left-rail>
+          <div class="h-full">
+            Left Rail
+          </div>
+        </template>
+        <template #right-rail>
+          <div class="h-full">
+            Right Rail
+          </div>
+        </template>
+      </p-layout-well-toolbar>
+    </template>
+  </ComponentPage>
+</template>
+
+<script lang="ts" setup>
+  import ComponentPage from '@/demo/components/ComponentPage.vue'
+</script>
+
+<style>
+.layout-well-section__box {
+  @apply
+  w-full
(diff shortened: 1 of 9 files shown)


src/styles/index.css
```diff
@@ -39,6 +39,7 @@
 
     /* Root Tokens */
     --p-color-divider: #51525C;
+    --p-color-divider-subdued: #51525C;
```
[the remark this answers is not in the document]
znicholasbrown: I also don't see this used anywhere, maybe worth removing?

src/layouts/PLayoutGrid/PLayoutGrid.vue
```diff
@@ -0,0 +1,110 @@
...
+    <main class="p-layout-grid__main">
+      <div v-if="$slots.leftRail" class="p-layout-grid__left-rail">
+        <slot name="leftRail" />
+      </div>
+      <div class="p-layout-grid__content">
+        <slot />
+      </div>
+      <div v-if="$slots.rightRail" class="p-layout-grid__right-rail">
+        <slot name="rightRail" />
+      </div>
```
znicholasbrown, line 32: If you follow the convention you set with the header and toolbar slots, this feels a little cleaner:
```suggestion
      <div v-if="$slots.mainLeading" class="p-layout-grid__main-leading">
        <slot name="mainLeading" />
      </div>
      <div class="p-layout-grid__content">
        <slot />
      </div>
      <div v-if="$slots.mainTrailing" class="p-layout-grid__main-trailing">
        <slot name="mainTrailing" />
      </div>
```


I'd also recommend using the `useSlots` composition Vue provides as it's generally recommended over the $ accessor.
```vue
<script lang="ts" setup>
import { useSlots } from 'vue'

const slots = useSlots()
</script>
```

And then updating your other slots from kebab -> pascal case just so they're internally consistent (this is one of those weird situations where kebab case is generally prefered but is gross when trying to access slots since you have to do `slots['kebab-case-slot']`

demo/sections/layouts/LayoutGrid.vue
```diff
@@ -0,0 +1,57 @@
...
+        <template #left-rail>
+          <div class="h-full">
+            Left Rail
+          </div>
+        </template>
+        <template #right-rail>
+          <div class="h-full">
+            Right Rail
+          </div>
+        </template>
```
znicholasbrown, line 41: These aren't showing up in the demo because your slots are named `leftRail` and `rightRail`, respectively

src/layouts/PLayoutGrid/PLayoutGrid.vue
```diff
@@ -0,0 +1,110 @@
...
+
+ <style>
+  .p-layout-grid {
+    @apply
+    w-full
+    h-full
+    border-l
+    flex
+    flex-col
+    border-[color:var(--p-color-divider)];
```
znicholasbrown, line 45: you should be able to do this instead for all of these:
```suggestion
    border-divider
```

src/layouts/PLayoutGrid/PLayoutGrid.vue
```diff
@@ -0,0 +1,110 @@
...
+    flex-1
+    w-full
+    relative
+    flex-nowrap;
+  }
+
+  .p-layout-grid__content {
+    @apply
+    flex-1
+    border-[color:var(--p-color-divider)];
```
znicholasbrown, line 97: This _might_ be intentional but I think the bug in the demo exposed that this border still shows up even when you've not provided rail slots:

image

znicholasbrown (commented):
Not blocking anything here since this is still in draft but hopefully these are helpful

znicholasbrown (approved):
thanks for the changes @aaazzam ! lgtm

PR #1145, opened 2024-03-12
Enhancement: `PMarkdownRenderer` memory update

znicholasbrown:
This PR moves all worker logic out of the `PMarkdownRender` component and into a singleton composition. This means a few things:

1. The renderer will be available for use outside of the component scope through the new `useMarkdownRenderer` composition
2. The package will only create a single worker instance shared by all renderers

This should resolve issues where creating many instances of `PMarkdownRenderer` caused browser tabs to crash due to browser limitations on the number of service workers and application could register. It should also speed up rendering in general since each instance after the first will have a worker ready.


znicholasbrown:
Note: I'd like to apply the same treatment to `PCode` now that the issue has been identified since it follows a similar pattern; it's getting a bit late so I'll tackle that first thing tomorrow morning.

znicholasbrown:
Ah there's an issue with reactivity of the initial text - investigating that

znicholasbrown:
Ok that should be fixed now, was accidentally passing the bare `props.text` instead of a getter

src/compositions/useMarkdownRenderer.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  callbacks.set(id, rendererCallback)
+  watchEffect(() => worker.postMessage({ id, text: toValue(text) }))
+
+  if (getCurrentScope()) {
+    onUnmounted(() => {
+      callbacks.delete(id)
+    })
+  }
```
pleek91, line 40: I THINK `onScopeDispose` is also a good way to do this. Doesn't depend on a component instance but still get disposed of when a component does.

pleek91 (approved):
Nice, this looks great

PR #1147, opened 2024-03-14
Enhancement: Small Base Inputs

aaazzam:
In this PR, PBaseInput - like PButton - can take a `small?: boolean` prop. 
- Follows the same pattern as PButton: explicitly passing small as true adds a computed class.
- This necessitated adding a `size?:` prop to PLoadingIcon to mirror our PIcon class. 

I've added the small size to the demo for easy viewing.

### DIFF
--- a/src/components/BaseInput/PBaseInput.vue
+++ b/src/components/BaseInput/PBaseInput.vue
@@ -4,7 +4,7 @@
       {{ prepend }}
     </div>
     <slot name="prepend" />
-    <slot name="control" :attrs="attrsWithDisabled" />
+    <slot name="control" :attrs="attrsForControl" />
     <div v-if="append" class="p-base-input__append">
       {{ append }}
     </div>
@@ -62,9 +62,13 @@
 
   const iconSize = computed(() => props.small ? 'small' : undefined)
 
-  const attrsWithDisabled = computed(() => ({
+  const attrsForControl = computed(() => ({
     ...attrs.value,
     disabled: props.disabled,
+    class: {
+      'p-base-input__control': true,
+      'p-base-input__control--small': props.small,
+    },
   }))
 </script>
 
@@ -90,7 +94,7 @@
   p-0
 }
 
-.p-base-input--small input { @apply
+.p-base-input__control--small { @apply
   py-1.5
   text-xs
   h-7
--- a/src/components/LoadingIcon/PLoadingIcon.vue
+++ b/src/components/LoadingIcon/PLoadingIcon.vue
@@ -10,17 +10,23 @@
 <script setup lang="ts">
   import { computed, defineProps } from 'vue'
 
-  const props = defineProps<{
-    size?: 'small' | 'large',
-  }>()
-  const iconSizeClass = computed(() => props.size && `p-loading-icon--${props.size}`)
+  const props = withDefaults(defineProps<{
+    size?: 'small' | 'default' | 'large',
+  }>(), {
+    size: 'default',
+  })
+
+  const iconSizeClass = computed(() => `p-loading-icon--${props.size}`)
 </script>
 
 <style>
 .p-loading-icon { @apply
   relative
-  w-4
-  h-4
+}
+
+.p-loading-icon--default { @apply
+  w-3
+  h-3
 }
 
 .p-loading-icon--small { @apply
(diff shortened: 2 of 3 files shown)


src/components/BaseInput/PBaseInput.vue
```diff
@@ -27,6 +29,7 @@
 </script>
 
 <script lang="ts" setup>
+  import { s } from 'vite/dist/node/types.d-jgA8ss1A'
```
brandonreid, line 32: Looks like a left over auto-imported something

brandonreid (approved):
One comment there, otherwise g2g!

src/components/BaseInput/PBaseInput.vue
znicholasbrown:
```suggestion
```
Looks like an auto-import snuck in

znicholasbrown (commented):
nice, only 1 comment code-wise

src/components/BaseInput/PBaseInput.vue
```diff
@@ -74,11 +81,22 @@
...
+  background-color: var(--p-color-input-bg);
   color: var(--p-color-input-text);
 }
 
+.p-base-input--small { @apply
+  text-xs
+  p-0
+}
+
+.p-base-input--small input { @apply
```
pleek91, line 93: My guess that this isn't enough. Base input is used with many different elements for the "input" such as `button` and `select`. 

Also increasing specificity here is going to cause problems because then a component implementing base-input will need a higher specificity in order to say changes its padding. 

I think a possible solution here would be to make a `p-base-input__control` block class and then add the modifier `p-base-input__control--small`. Then both classes can be added to what is currently called `attrsWithDisabled` which could probably get renamed to `attrsForControl` or something.

znicholasbrown (approved):

src/components/LoadingIcon/PLoadingIcon.vue
```diff
@@ -1,19 +1,38 @@
...
     <div class="p-loading-icon__segment" />
   </div>
 </template>
 
+<script setup lang="ts">
+  import { computed, defineProps } from 'vue'
+
+  const props = defineProps<{
+    size?: 'small' | 'large',
+  }>()
```
pleek91, line 15: should we be consistent here and just add an optional `small` prop? I don't feel strongly, just crossed my mind.

src/components/LoadingIcon/PLoadingIcon.vue
```diff
@@ -1,19 +1,38 @@
...
+  const iconSizeClass = computed(() => props.size && `p-loading-icon--${props.size}`)
```
pleek91, line 16: Let's not use boolean logic so we don't end up trying to bind `false` as a class. Should be fine to just always add the class as long as there is a default. 
```suggestion
  const iconSizeClass = computed(() => `p-loading-icon--${props.size ?? 
  'large'}`)
```

pleek91:
@aaazzam just lol that all three of use commented about that import 😂

src/components/LoadingIcon/PLoadingIcon.vue
pleek91: TY 🙏

src/components/LoadingIcon/PLoadingIcon.vue
pleek91: Up to you, this ultimately doesn't matter much other than just improving the devx/consistency. We could never change it and probably wouldn't matter lol.

pleek91: Yeah don't feel like you need to address this. 

I don't really have a preference. Individual props are kinda nice at implementation, but they can get weird in component code. Either is fine by me, so if we do unify these maybe we just pick whichever one is already used the most so we have the least amount of changes to make.

src/components/LoadingIcon/PLoadingIcon.vue
```diff
@@ -1,17 +1,42 @@
...
+
+.p-loading-icon--default { @apply
+  w-3
+  h-3
+}
+
+.p-loading-icon--small { @apply
+  w-3
+  h-3
+}
```
pleek91, line 35: Why have default if its the same as small? Could just default the size prop to `small` yeah? Maybe this is copy pasted from another component, so don't take this question personally lol.

pleek91 (approved):

PR #1150, opened 2024-03-14
Enhancement: Small Select

aaazzam:
In this PR, PSelect - like PBaseInput - can take a small?: boolean prop.

Follows the same pattern as PBaseInput: explicitly passing small as true adds a computed class.
- This necessitated adding a size?: prop to Checkbox.

I've added the small size to the demo for easy viewing.

### DIFF
--- a/src/components/Checkbox/PCheckbox.vue
+++ b/src/components/Checkbox/PCheckbox.vue
@@ -1,5 +1,5 @@
 <template>
-  <PLabel class="p-checkbox" :class="classes" :style="attrStyles">
+  <PLabel class="p-checkbox" :class="classes.checkbox" :style="attrStyles">
     <template v-if="slots.label || isDefined(label)" #label>
       <slot name="label">
         <span class="p-checkbox__label">
@@ -15,6 +15,7 @@
         type="checkbox"
         :disabled="disabled"
         class="p-checkbox__input"
+        :class="classes.input"
       >
     </template>
   </PLabel>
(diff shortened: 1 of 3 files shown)


src/components/Select/PSelectButton.vue
```diff
@@ -60,6 +64,13 @@
   cursor-not-allowed
 }
 
+.p-base-input__control--small{
+  @apply
+  py-1
+  pl-1.5
```
znicholasbrown, line 70: I think it'd be fine if this were `px-2`, personally

znicholasbrown (approved):
code looks about what i'd expect, thanks @aaazzam !

src/components/Checkbox/PCheckbox.vue
```diff
@@ -81,6 +83,11 @@
   w-auto
 }
 
+.p-checkbox--small .p-checkbox__input { @apply
+  h-3
+  w-3
+}
```
pleek91, line 89: The checkbox itself gets smaller? They're pretty small as is aren't they? 

Also `p-checkbox__input--small`

src/components/Select/PSelectButton.vue
```diff
@@ -34,6 +34,10 @@
   const el = computed(() => buttonElement.value)
 
   defineExpose({ el, wrapper })
+
+  const props = defineProps<{
+    small?: boolean,
```
pleek91, line 39: This is because your defining `const props` but not using `props.small` in the script. And in the template you shouldn't use `props.small` just `small`

pleek91:
I get making the select itself smaller to match the other small inputs. But the dropdown probably could be normal size yeah? Looks very squished IMO

Take or leave that feedback. No response necessary

image

pleek91 (approved):
Thanks for answering questions!

PR #1151, opened 2024-03-14
Indeterminate `<p-checkbox />` styles

collincchoy:
This PR enhances the `<p-checkbox />` component to better handle [indeterminate state checkboxes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes). This PR adds a new section to the checkbox demo - check it out there. Below is a static version of it:

| Description | 📸 from Demo |
| ---- | ---- |
| None selected | image |
| Some selected | image |
| All selected | image |

At the moment like a lot of the other checkbox styles, the indeterminate state styles are coming from from the tailwindcss-forms plugin (https://github.com/tailwindlabs/tailwindcss-forms/blob/e6bcd1ff718ad2d0380a65ded0564369e4cbda32/src/index.js#L291C3-L316C9) however as-is, they clash with some of our additional styles and also don't conform well to dark/light mode. 

| Mode | **Before** | **After** |
| ---- | ---- | ---- |
| 🌑 | image | image |
| 🌕 | image | image |

### DIFF
--- a/src/components/Checkbox/PCheckbox.vue
+++ b/src/components/Checkbox/PCheckbox.vue
@@ -97,8 +97,8 @@
   focus:ring-offset-focus-ring
   focus:ring-offset-focus-ring-offset;
   background-color: var(--p-color-input-bg);
+  color: var(--p-color-input-bg);
   border-color: var(--p-color-input-border);
-  color: var(--p-color-input-text);
 }
 
 .p-checkbox__input:focus:not(:focus-visible) { @apply
@@ -107,17 +107,20 @@
 }
 
 .p-checkbox__input:not(:disabled):hover {
-  background-color: var(--p-color-input-bg);
+  color: var(--p-color-input-bg);
   border-color: var(--p-color-input-border-focus);
-  color: var(--p-color-input-text);
 }
 
 .p-checkbox__input:checked,
 .p-checkbox__input:checked:hover,
 .p-checkbox__input:checked:focus {
-  background-color: var(--p-color-input-checked-bg);
+  color: var(--p-color-input-checked-bg);
   border-color: var(--p-color-input-checked-border);
-  color: var(--p-color-input-checked-text);
+}
+
+.p-checkbox__input:indeterminate {
+  border-color: var(--p-color-input-border);
+  background-image: var(--p-color-input-checkbox-indeterminate-bg-img);
 }
 
 .p-checkbox--failed {
@@ -127,7 +130,7 @@
 .p-checkbox--failed .p-checkbox__input,
 .p-checkbox--failed .p-checkbox__input:hover,
 .p-checkbox--failed .p-checkbox__input:focus {
-  background-color: var(--p-color-input-bg-invalid);
+  color: var(--p-color-input-bg-invalid);
   border-color: var(--p-color-input-border-invalid);
 }
 
--- a/src/components/Checkbox/PCheckbox.vue
+++ b/src/components/Checkbox/PCheckbox.vue
@@ -120,7 +120,7 @@
 
 .p-checkbox__input:indeterminate {
   border-color: var(--p-color-input-border);
-  background-image: var(--p-color-input-checkbox-indeterminate-bg-img);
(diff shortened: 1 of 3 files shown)


pleek91 (approved):
Love the `indeterminate` attribute! LGTM

src/components/Checkbox/PCheckbox.vue
```diff
@@ -97,8 +97,8 @@
   focus:ring-offset-focus-ring
   focus:ring-offset-focus-ring-offset;
   background-color: var(--p-color-input-bg);
```
collincchoy, line 99: Open to reverting, but something I noticed whilst digging through this:
these `color` rules don't do anything since the "check"(:checked) and the "minus"(:indeterminate) are background-image SVGs from the tailwind forms plugin. 

We could certainly _make_ these work with some more tweaks, but atm they don't. 

Additionally, the tailwindcss-forms base styles are designed to support setting the background color via utilities like `text-blue-500` so the different pseudo selectors are styled like `[type='checkbox']:checked { background-color: currentColor }` which means that for each state, we don't have to override the background-color. Instead, we can just set the color (similar to using a `text-blue-500` as is in their docs). 

So here, I'm removing the background-color overrides in favor of just setting via `color`. I think this also maps more closely to newer checkbox styling via `accent-color`. 


That said, I could see that it might be a little confusing this way that `color` is what's ultimately influencing the "background-color"...

src/styles/index.css
```diff
@@ -275,6 +277,8 @@
     --p-color-input-checked-border: var(--p-color-selected);
     --p-color-input-checked-text: var(--p-color-text-default);
 
+    --p-color-input-checkbox-indeterminate-bg-img: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' fill='' viewBox='0 0 16 16'%3e%3cpath stroke='%2370707B' stroke-linecap='round' stroke-linejoin='round' stroke-width='2' d='M4 8h8'/%3e%3c/svg%3e");
```
collincchoy, line 280: The difference here is having the lightmode use a different stroke color - here it's using `%2370707B` which is a url encoded hex `#70707B` which is our `--p-color-input-text` in light mode. 

effectively doing what the before-my-changes `color: var(--p-color-input-text)` was intending to do.

src/styles/index.css
```diff
@@ -105,6 +105,8 @@
     --p-color-input-checked-border: var(--p-color-selected);
     --p-color-input-checked-text: var(--p-color-text-default);
 
+    --p-color-input-checkbox-indeterminate-bg-img: url("data:image/svg+xml,%3csvg xmlns='http://www.w3.org/2000/svg' fill='' viewBox='0 0 16 16'%3e%3cpath stroke='white' stroke-linecap='round' stroke-linejoin='round' stroke-width='2' d='M4 8h8'/%3e%3c/svg%3e");
```
collincchoy, line 108: this svg data url is directly from tailwindcss-forms which is the current base value

https://github.com/tailwindlabs/tailwindcss-forms/blob/e6bcd1ff718ad2d0380a65ded0564369e4cbda32/src/index.js#L296

I've pulled it in so that I could recolor it for lightmode since in its current form here, the stroke is `white` which blends in with the already-white background of the checkbox in lightmode making it IMPERCEIVABLE.

collincchoy, line 108: is this really weird that this is prefixed with `--p-color`... 

this feels like an appropriate place to put this but every single variable in this area is prefixed with `--p-color` 😶‍🌫️

src/styles/index.css
collincchoy: sucks that i can't pull in a variable for this. 


and a little sucks also that this isn't immediately obvious that it's using a color from a variable (like via cmd+F searching for the hex value)

😞

src/styles/index.css
pleek91: I don't think it needs to color prefix. 🤷

pleek91 (approved):

PR #1154, opened 2024-03-15
Feature: `PContextMenu`

znicholasbrown:
This PR implements a more generic and extended version of `PIconButtonMenu`. It exposes slots for `target` and `context-menu` (overrides the default `POverflowMenu` for maximum customization). The `default` re-exposes the `default` slot of `POverflowMenu`.

Some more interesting enhancements are the addition of behavior to open the context menu at position of the mouse. This should be helpful when creating true context menus (e.g. overriding the browser right-click behavior) or when position around the target container isn't quite the right behavior.

Check out the demo for some working samples of both


znicholasbrown:
Ah we've got a demo that's bulk importing everything from the positions utility file... hmm I'll have to filter these out somehow

src/compositions/useMousePosition.ts
pleek91, line 1: Two things here. 

1. This would be great in vue-compsitions
2. Might be worth it to only attach the listeners if something is actually using it since mousevents are going to happen pretty much all the time.

src/types/position.ts
```diff
@@ -1,6 +1,11 @@
 import { Ref } from 'vue'
 import { Pixels } from '@/types/units'
 
+export type MousePosition = {
+  x: number,
+  y: number,
+}
+
```
pleek91, line 8: nit: I'd put this in useMousePosition.ts since its meant to be used with that. Its not really the same type of position as these other types.

src/utilities/positionRelative.ts
pleek91, line 1: Nice use of factories to create popover placements!

src/components/ContextMenu/PContextMenu.vue
```diff
@@ -0,0 +1,90 @@
...
+    if (props.disableCloseOnEscape) {
+      return
+    }
+
+    if (isKeyEvent(keys.escape, event)) {
+      internalClose()
+    }
+  }
+
+  useGlobalEventListener('keyup', closeOnEscape)
```
pleek91, line 83: If you don't care specifically about keyup vs keydown there's a `useKeyDown` composition in vue-compositions that abstracts away the "which key" part pretty nicely. Realizing there should be a `useKeyUp` composition as well.

src/utilities/positionRelative.ts
```diff
@@ -0,0 +1,46 @@
...
+export const lowerRight = function(position: MaybeRefOrGetter<MousePosition>): PositionMethod {
+  return function(): Position {
+    const { x, y } = toValue(position)
+
+    return {
+      top: y,
+      left: x,
+    }
+  }
+}
```
pleek91, line 13: Could these be made more generic and just use the built in placement methods we already have but override the target to mouse position?

src/components/ContextMenu/PContextMenu.vue
```diff
@@ -0,0 +1,90 @@
...
+  import PPopOver from '@/components/PopOver/PPopOver.vue'
+  import { useMousePosition } from '@/compositions'
+  import { PositionMethod, keys } from '@/types'
+  import { isKeyEvent } from '@/utilities'
+  import { bottomLeft, bottomRight, topLeft, topRight } from '@/utilities/position'
+  import { lowerLeft, lowerRight, upperLeft, upperRight } from '@/utilities/positionRelative'
+
+  const props = withDefaults(defineProps<{
+    placement?: PositionMethod | PositionMethod[],
+    autoClose?: boolean,
```
pleek91, line 35: should this be `disableAutoClose` since the other close props are true by default?

src/compositions/useMousePosition.ts
znicholasbrown: I thought you might ask for that, getting that PR ready haha

znicholasbrown: Ok that's [here](https://github.com/PrefectHQ/vue-compositions/pull/404)

To your second point, that's tackled in the vue-compositions PR

src/components/ContextMenu/PContextMenu.vue
znicholasbrown: Mmm yeah you're right, I originally was just re-exposing the `PPopover` prop but I think this makes more sense

src/components/ContextMenu/PContextMenu.vue
znicholasbrown: Very nice suggestion, updated

src/utilities/positionRelative.ts
znicholasbrown: Ok what do you think about now?

pleek91: Nice, beautiful

pleek91 (approved):

PR #1160, opened 2024-03-18
Enhancement: `PCodeHighlight` singleton worker

znicholasbrown:
This PR moves worker logic out of the `PCodeHighlight` component and into a composition that uses a singleton worker. Similar to #1145, functionality remains identical but memory usage should be drastically more efficient when mounting many of these components at the same time and highlight latency should also be improved.

### DIFF
--- a/src/types/codeHighlight.ts
+++ b/src/types/codeHighlight.ts
@@ -29,16 +29,18 @@ export function isSupportedLanguage(lang: unknown): lang is SupportedLanguage {
 }
 
 export type UnformattedMessagePayload = {
+  id: string,
   text: string,
   lang: SupportedLanguage,
 }
 
 export type FormattedMessagePayload = {
-  unformatted: string,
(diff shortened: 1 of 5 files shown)


src/types/codeHighlight.ts
```diff
@@ -29,16 +29,18 @@ export function isSupportedLanguage(lang: unknown): lang is SupportedLanguage {
...
 
 export type FormattedMessagePayload = {
-  unformatted: string,
+  id: string,
+  unformatted?: string | undefined,
   formatted: string,
   illegal: boolean,
   relevance: number,
-  language: string,
+  language?: string | undefined,
```
znicholasbrown, line 43: These types were technically incorrect before but we're not using the return values anywhere in the component and the worker message wasn't typed

pleek91 (approved):
Nice, love the `useCodeHighlight` composition.

PR #1162, opened 2024-03-19
Add p-select-all-checkbox

collincchoy:
This PR adds a new component `<p-select-all-checkbox />`. It wraps a `p-checkbox` to implement the following pattern:
- when nothing is selected, the checkbox is unchecked
- when some-but-not-all are selected, the checkbox is checked and visually appears in an indeterminate state
- when all selectable items are selected, the checkbox is checked and visually appears in a checked state

- on check, all items should be selected
- on uncheck, all items should be deselected.

The component contains a tooltip along with a sr-friendly label.

This PR updates `<p-table />` to use `<p-select-all-checkbox />` in-place of its prior `<p-checkbox />`

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -15,7 +15,8 @@
   "globals": {
     "defineProps": "readonly",
     "defineEmits": "readonly",
-    "defineExpose": "readonly"
+    "defineExpose": "readonly",
+    "defineModel": "readonly"
   },
   "settings": {
     "import/resolver": {
--- a/demo/sections/components/Checkbox.vue
+++ b/demo/sections/components/Checkbox.vue
@@ -4,7 +4,7 @@
     :demos="[
       { title: 'Checkbox' },
       { title: 'Multiple', description: 'For multiple checkboxes, the v-model value can be an array to be shared across all checkboxes. See also: p-checkbox-group' },
-      { title: 'Indeterminate', description: 'A checkbox can either be on or off, but there is a third, visual-only state called indeterminate. This is used to indicate that a checkbox is in a mixed state, for example, when a parent checkbox is checked, but not all of its children are - e.g. a `Select All` checkbox.' },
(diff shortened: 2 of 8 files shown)


znicholasbrown (approved):
This looks great @collincchoy; 1 additional thought; the [SO question](https://ux.stackexchange.com/questions/92070/should-clicking-a-checkbox-in-an-indeterminate-state-check-it-or-clear-it-why) where ~your~the checkbox demo comes from highlights behavior I think we might want to replicate, which is that interacting with a select-all checkbox in an indeterminate state should evaluate to `true` instead of `false` for child items. Currently this is doing the opposite (which I think was what the `PTable` component's checkbox behavior was previously as well so I understand the consistency). How would you feel about flipping that behavior?

znicholasbrown:
I just realized you didn't introduce that demo so updated my response to include a link, sorry about that!

znicholasbrown:
Discussed the behavior change sync - all clear here 👍🏻

znicholasbrown (approved):

collincchoy:
> This looks great @collincchoy; 1 additional thought; the [SO question](https://ux.stackexchange.com/questions/92070/should-clicking-a-checkbox-in-an-indeterminate-state-check-it-or-clear-it-why) where ~your~the checkbox demo comes from highlights behavior I think we might want to replicate, which is that interacting with a select-all checkbox in an indeterminate state should evaluate to `true` instead of `false` for child items. Currently this is doing the opposite (which I think was what the `PTable` component's checkbox behavior was previously as well so I understand the consistency). How would you feel about flipping that behavior?

For the archives:
decided to go this direction because 
A) doesn't seem to be a clear general consensus for whether clicking whilst in a partially selected state should select all or deselect all 
B) collin thinks the minus sign implies clicking would remove
C) GMail's select all checkbox goes this direction so there's some precedent/muscle-memory to lean on

PR #1175, opened 2024-03-29
Feature: `PCascadePanels`

znicholasbrown:
This PR introduces a new set of utility components/compositions, called Cascade Panels. These components are intended to allow us to create flexible nested menu experiences without the overhead of managing their state; while this PR is large, most of that is demo, the actual components/compositions themselves should be fairly straightforward. 

While I'd like to merge/release these, I'm still working on live versions so may revisit some functionality. One dev ex piece I'd already like to include (but won't in this PR) is for panel components to emit `close` events that callers can listen to, to do something like: `@close="openPanelById('some-next-panel')` without needing to either listen to a nested v-model or watch a panel.


src/compositions/cascade/useCascadePanel.ts
```diff
@@ -0,0 +1,53 @@
...
+    throw new Error('Cascade panel not found. Are you sure the component calling useCascadePanel() exists within a <p-cascade-panel> component context?')
+  }
+
+  return cascadePanel
+}
+
+export function useCascadePanel(id?: MaybeRefOrGetter<CascadePanelId>): UseCascadePanel {
+  if (!id) {
+    return getInjectedPanel()
+  }
```
pleek91, line 26: What's the use case for using this with or without an id?

src/compositions/cascade/useCascadePanel.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  const cascadePanel: UseCascadePanel = {
+    isOpen,
+    toggle,
+    open,
+    close,
+  }
+
+  provide(id, cascadePanel)
```
pleek91, line 51: You're providing a `MaybeRefOrGetter` as the key which probably causes some weird things to happen when attempting to inject this. Guessing you meant to provide it with the actual value of the id? I'm surprised typescript isn't complaining since I think the key can only be a string or a symbol. 

Though that kinda contradicts the fact that `MaybeRefOrGetter` implies reactivity. Meaning that if the id changes everything in this composition should change. But `provide` is not reactive. 

Further makes me think this doesn't actually work like you're thinking.

src/compositions/cascade/useCascadePanels.ts
```diff
@@ -0,0 +1,178 @@
...
+
+export function getInjectedCascadePanels(): UseCascadePanels {
+  const cascadePanels = inject(cascadePanelsKey)
+
+  if (!cascadePanels) {
+    throw new Error('Cascade panels not found. Are you sure the component calling useCascadePanels() exists within a <p-cascade> component context?')
+  }
+
+  return cascadePanels
+}
```
pleek91, line 22: IMO its a little bit of a red flag that this is an exported utility but it must be run in a component context. I'd personally rename this to be a composition so that there's no confusion.

src/compositions/cascade/useCascadePanels.ts
```diff
@@ -0,0 +1,178 @@
...
+  if (!panelsRefOrGetter) {
+    try {
+      return getInjectedCascadePanels()
+    } catch {
+      throw new Error('No panels were provided; and no cascade panel context was found. Are you sure the component calling useCascadePanels() exists within a <p-cascade> component context?')
+    }
+  }
+
+  const panels = toRef(panelsRefOrGetter)
+  const state = reactive<CascadeState>(Object.fromEntries(panels.value.map((panel) => [panel.id, false])))
```
pleek91, line 65: `panelsRefOrGetter` is reactive data but its being used in a non reactive way here. If that value changes this state object won't update to reflect any panels that were added or removed.

src/compositions/cascade/useCascadePanel.ts
znicholasbrown: Ah nope, that was bad leftover from an earlier version, this wasn't intended to be reactive at all and I ended up poorly mixing the two versions. That should be fixed now, great catch

src/compositions/cascade/useCascadePanels.ts
znicholasbrown: Ah we don't need to be doing this at all - updated to reflect that. The state is really only important for tracking if panels are open. If panels exist but have never been opened, they don't need to be in the `state` object yet but this made it look like they did

src/compositions/cascade/useCascadePanels.ts
znicholasbrown: Hm that's fair, it's only really meant for internal use (of these components) since `useCascade*` compositions return this already. That said, I agree that's a little confusing so I renamed; I could remove the export altogether in favor of calling the injection directly in the `useCascadePanel` composition (that's why this is exported in the first place) - I'd want to duplicate the error handling though, which isn't a huge deal 🤷🏻

pleek91: If it was me I would duplicate the logic but create an error class and throw that error. Not exporting the utility has more pros than the loss of dryness.

znicholasbrown: Works for me then - updated 👍🏻

src/compositions/cascade/useCascadePanels.ts
```diff
@@ -0,0 +1,177 @@
...
+    } catch {
+      throw new PanelsNotProvided()
+    }
+  }
+
+  const panels = toRef(panelsRefOrGetter)
+  const state = reactive<CascadeState>({})
+  const isOpen = ref(false)
+  const openPanels = computed(() => panels.value.filter((panel) => state[panel.id]))
+  const panelIsOpen = computed(() => (id: CascadePanelId) => state[id])
```
pleek91, line 70: I don't think this needs to be reactive does it? `state` is reactive so if this was just a function you could use it inside another computed or in a template. I'd just make this a function like `getPanelIsOpenById` like the other functions in this composition.

src/components/CascadeMenu/PCascadePanels.vue
```diff
@@ -0,0 +1,102 @@
...
+    return `${kebabCase(panelId.toString())}`
+  }
+</script>
+
+<style>
+.p-cascade-panels { @apply
+  flex
+  flex-wrap
+  gap-2
+  pointer-events-none
```
pleek91, line 75: Curious why this has `pointer-events-none`

demo/components/cascade/AdvancedDemo.vue
```diff
@@ -0,0 +1,131 @@
...
+    {
+      id: 'borders',
+      level: 2,
+    },
+  ]
+
+  const { openPanelById, closePanelById, toggle, close } = useCascadePanels(panels)
+
+  const placement = ref([positions.bottomLeft])
+  const { styles, target, container, content } = useMostVisiblePositionStyles(placement)
```
pleek91, line 79: So the panels components and compositions manage if something is visible or not? There's not actually any built in functionality for opening or closing based on interaction. It just handles visibility and layout?

pleek91: discussed sync, might follow up and bring the placement functionality into the panels component.

src/components/CascadeMenu/PCascadePanels.vue
```diff
@@ -0,0 +1,88 @@
...
+    closePanelById,
+    getPanelById,
+    isOpen,
+    open,
+    openPanelById,
+    openPanels,
+    panelIsOpen,
+    toggle,
+    togglePanelById,
+  } = inject(cascadePanelsKey, undefined) ?? useCascadePanels(() => props.panels)
```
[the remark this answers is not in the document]
znicholasbrown: It's more a devex piece so that you can do something like:
```vue
<template>
  <p-cascade-panels :panels />
</template>

<script lang="ts" setup>
import { useCascadePanels } from '@prefecthq/prefect-design'

const panels = [/* whatever */]

const { open, openPanelById, ...rest } = useCascadePanels(panels)
</script>
```
without needing to bind the component like we do elsewhere:

```vue
<template>
  <p-cascade-panels ref="cascadePanels" :panels />
</template>

<script lang="ts" setup>
import { ref } from 'vue'
import { PCascadePanels } from '@prefecthq/prefect-design'

const panels = [/* whatever */]

const cascadePanels = ref<typeof PCascadePanels>()
// now we can use the same methods via cascadePanels.value.method but we need to check
// everywhere that cascadePanels.value exists

</script>

src/compositions/cascade/useCascadePanel.ts
znicholasbrown: This is another devex piece - so the `PCascadePanel` component wrapper can call `useCascadePanel(panelId)`, which will `provide` the hydrated panel without needing to passing around ids. Then in a downstream component you can do `useCascadePanel` without any further injection/knowledge of the id

src/components/CascadeMenu/PCascadePanels.vue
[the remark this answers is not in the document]
pleek91: So if `useCascadePanels` has already been called at a higher level that same context will be used in the panels component otherwise it creates its own context?

src/components/CascadeMenu/PCascadePanels.vue
znicholasbrown: There might be another way to handle this but I found in testing that there was often a big difference in the heights/widths of panels which caused the bounding box of this component to be massive. Unless we ignore pointer events on this (you'll see this is reset on the panels themselves), the bounding box invisibly captures click events when using this in an absolutely-positioned context

src/components/CascadeMenu/PCascadePanels.vue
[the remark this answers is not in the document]
znicholasbrown: Correct!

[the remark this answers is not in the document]
znicholasbrown: Though I wonder if we might get into trouble with that and this needs to check if the panels are equivalent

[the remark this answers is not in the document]
znicholasbrown: Actually yes we do need to check that lol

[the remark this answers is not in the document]
pleek91: lol I trust your instincts here, just wanted to make sure I understood the intent

src/compositions/cascade/useCascadePanels.ts
znicholasbrown: Ok this makes sense - updated that

src/components/CascadeMenu/PCascadePanels.vue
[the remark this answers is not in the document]
znicholasbrown: Ok I added some logic to handle injection scenarios like this; it's not the prettiest but I think the devX is nicer for it

pleek91 (approved):

PR #1182, opened 2024-04-02
[p-select] Fix multi select keyboard selection

collincchoy:
`<p-select multiple ... />` components and others that build on top if it (i.e. comboboxes) have a bug when attempting to select options via keyboard. Pressing <kbd>Enter</kbd> doesn't do anything like it should once an option is highlighted. 

### Steps to reproduce:
1. Go to the demo (on main) https://prefect-design.netlify.app/components/select#multi-select
2. Pop it open
3. Press down arrows to navigate to an option
4. Press <kbd>Enter</kbd> but nothing happens

### DIFF
--- a/demo/sections/components/Combobox.vue
+++ b/demo/sections/components/Combobox.vue
@@ -106,11 +106,9 @@
 </script>
 
 <style>
-  .combobox__demo {
-    @apply
-    h-80
-    flex
-    flex-col
-    gap-4
-  }
+.combobox__demo { @apply
+  flex
+  flex-col
+  gap-4
+}
 </style>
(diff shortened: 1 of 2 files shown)


pleek91 (approved):
Didn't test. Assume it works like you said lol

demo/sections/components/Combobox.vue
```diff
@@ -106,11 +106,9 @@
...
-    h-80
-    flex
-    flex-col
-    gap-4
-  }
+.combobox__demo { @apply
+  flex
+  flex-col
+  gap-4
+}
```
collincchoy, line 113: I removed `h-80` from here. 

| **Before** | **After** |
| ---- | ---- |
| image | image |

Why? 
I didn't even notice there were more demos further down that i had to scroll to.. also more consistent w/ p-select demo page.

PR #1183, opened 2024-04-03
Bugfix: Don't bind panel to slot

znicholasbrown:
I'm not quite sure why this is necessary but after spending all evening chasing down an HMR-only recursion issue the culprit seems to be the fact that I'm binding the `panel` prop in both the `PCascadePanel` component and again in the named slot in `PCascadePanels` loop. 

Note: This caused me to change an earlier demo a bit - instead of accessing the panel through the binding it's calling `closePanelById` - downstream components are fine because they can still access the panel via the `usePanel` composition, which is preferred anyway.

### DIFF
--- a/demo/components/cascade/BasicDemo.vue
+++ b/demo/components/cascade/BasicDemo.vue
@@ -12,10 +12,10 @@
       <Level1 />
     </template>
 
-    <template #level-2="{ close }">
+    <template #level-2>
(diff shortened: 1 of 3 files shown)


pleek91 (approved):

znicholasbrown:
Hm when using this locally the issue started appearing again so I'm going to close this for now... 😕

PR #1199, opened 2024-04-11
Bugfix: `PTagWrapper` justify

znicholasbrown:
https://github.com/PrefectHQ/prefect-design/pull/1055 fixed some issues with tag wrapper containers without explicit widths. However it broke functionality that allowed the caller to justify content in any direction and still hide children as expected. This PR does a few things to address that as well as improve QOL of the component in general:
- the wrapper now adds `justify-*` classes when the `justify` prop is passed
- inline width calculation now takes into account the calculated column gap
- the resize observer is now throttled with a trailing edge calculation - this should both improve performance and fix race conditions the first time the observer is run
- the `calculateOverflow` method is now part of the component's exposed methods; this allows callers to manually re-evaluate overflow as needed in cases where tags either aren't being passed or tag content is being loaded asynchronously

Bonus:
- fixes issues with the demo resize component not working properly 
- adds demos for center, left, and right justifications

### DIFF
--- a/demo/components/ResizableSection.vue
+++ b/demo/components/ResizableSection.vue
@@ -7,7 +7,7 @@
     </p-frame>
 
     <div class="resizable-section__aside">
-      <div class="resizable-section__handle" @mousedown="start" @mouseup="stop">
+      <div class="resizable-section__handle" @mousedown="start">
         <component :is="ResizeIcon" />
       </div>
 
@@ -23,6 +23,7 @@
 <script lang="ts" setup>
   import { useColorTheme } from '@/compositions'
   import { toPixels } from '@/utilities'
+  import { useGlobalEventListener, useMousePosition } from '@prefecthq/vue-compositions'
   import { computed, ref, nextTick } from 'vue'
   import ResizeIcon from '@/demo/components/ResizeIcon.svg'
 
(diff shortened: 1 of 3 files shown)


src/components/TagWrapper/PTagWrapper.vue
```diff
@@ -223,25 +271,13 @@
...
-  pl-1;
-}
-
-.p-tag-wrapper__tag--center { @apply
-  px-[0.125rem];
-}
-
-.p-tag-wrapper__tag--left { @apply
-  pr-1;
-}
```
znicholasbrown, line 243: These were arbitrary - we now rely on the `gap` attr

demo/sections/components/Tags.vue
```diff
@@ -71,7 +75,7 @@
         <div>
           <p>Inline</p>
           <div class="border border-default p-2 max-w-full">
-            <p-tag-wrapper :tags="numberArr" justify="left" inline class="border border-default" />
+            <p-tag-wrapper :tags="numberArr" justify="left" inline class="border border-default test-123" />
```
pleek91, line 78: Guessing `test-123` is something left for debugging

src/components/TagWrapper/PTagWrapper.vue
```diff
@@ -161,17 +188,31 @@
...
+      const boundingBox = tag.getBoundingClientRect()
 
-    container.value!.style.width = `${totalTagsWidth}px`
+      return acc + boundingBox.width
+    }, 0))
   }
 
   function createObserver(): void {
-    resizeObserver = new ResizeObserver(calculateOverflow)
+    const throttledCalculateOverflow = throttle(calculateOverflow, 100, { trailing: true })
```
pleek91, line 214: I'm surprised we need `throttle`. A resize observer should be tuned to the animation frame. Was that just to fast?

demo/components/ResizableSection.vue
pleek91, line 1: Ahh this actually works nicely now! ❤️

pleek91 (approved):
Buttery smooth from my testing. Nicely done

src/components/TagWrapper/PTagWrapper.vue
znicholasbrown: Yeah, justttttt a hair too fast; logging the container width was the biggest giveaway - it was always 0 the first time this ran without a throttle

demo/sections/components/Tags.vue
znicholasbrown: Ah yes nice catch

PR #1202, opened 2024-04-12
Bugfix: `PTagWrapper` with async data

znicholasbrown:
This PR adds a mutation observer to better track overflow when down-DOM updates occur, specifically in cases of asynchronously-loaded data.

Before:

https://github.com/PrefectHQ/prefect-design/assets/27291717/49f637a2-ab04-4c6c-b1ba-6bd7ae9cb15c

After:

https://github.com/PrefectHQ/prefect-design/assets/27291717/7edd2178-fd82-4b14-9db6-1cd5c748a16b

### DIFF
--- a/src/components/TagWrapper/PTagWrapper.vue
+++ b/src/components/TagWrapper/PTagWrapper.vue
@@ -4,7 +4,7 @@
       <template v-for="tag in tags" :key="tag.value">
         <div class="p-tag-wrapper__tag" :class="classes.tag">
           <slot name="tag" :tag="tag">
-            <PTag :value="tag" />
+            <PTag :value="tag" :small="small" />
           </slot>
         </div>
       </template>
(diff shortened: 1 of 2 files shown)


src/components/TagWrapper/PTagWrapper.vue
```diff
@@ -4,7 +4,7 @@
       <template v-for="tag in tags" :key="tag.value">
         <div class="p-tag-wrapper__tag" :class="classes.tag">
           <slot name="tag" :tag="tag">
-            <PTag :value="tag" />
+            <PTag :value="tag" :small="small" />
```
znicholasbrown, line 7: Also noticed this wasn't being passed even though it is in most upstream components

collincchoy (approved):
tested w/ p-select locally and seems to resolve the issue 👍

PR #1203, opened 2024-04-13
Enhancement: Add a footer slot to the wizard component

zhen0:
We want to add the mini event feed to the bottom of the trigger step in the automation wizard.  This PR adds a footer slot after the actions slot in the wizard to facilitate that.  It also adds an emit of the current step so that we can easily identify which step the wizard is on and conditionally show a footer. 

image

### DIFF
--- a/src/components/Wizard/PWizard.vue
+++ b/src/components/Wizard/PWizard.vue
@@ -21,7 +21,7 @@
 
       <slot name="after-steps" />
 
-      <div class="p-wizard__footer">
+      <div class="p-wizard__actions">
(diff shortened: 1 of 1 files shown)


src/components/Wizard/PWizard.vue
```diff
@@ -36,6 +36,7 @@
           </p-button>
         </slot>
       </div>
+      <slot name="footer" />
```
pleek91, line 39: You could use the same logic from the content slot and have the footer be dynamic based on the step. That way you could target the footer for a specific step.

pleek91 (approved):
One thought but LGTM

PR #1209, opened 2024-04-16
[p-wizard] Skip navigations

collincchoy:
This PR adds navigation "shortcuts" to the wizard namely:
- step headers are now clickable
  - if you've completed the step already, then you can go straight to that step via headers rather than stepping 1-by-1
  - you can pass the `nonlinear` prop to the wizard to allow skipping steps without the completion check - i.e. when revisiting a previously completed wizard form
- when the wizard is passed a `show-save-and-exit` prop, a new button is available on all but the last step to execute the current step's validator and then emit a 'submit' event without having to step through the entire wizard. This gives you a short circuit out of the wizard - i.e. when revisiting a previously completed wizard form and attempting to make a change in 1 step without having to cycle through the rest of the wizard to submit.


pleek91 (approved):
LGTM!

PR #1210, opened 2024-04-16
[p-combobox] Remove quotes around unknown combobox options/values.

collincchoy:
| **Before** | **After** |
| ---- | ---- |
| image | image |
| image | image |


pleek91 (approved):
Yeah I think the quotes were over ambitions. I think the idea was that you could differentiate between known values and custom values. But doesn't seem necessary. LGTM!

PR #1214, opened 2024-04-17
Hide PModal footer when no actions are passed

aaazzam:
Currently our Modal component creates a "Cancel" action by default even if no actions are passed. Since action-less modals are used for previews, there's no action to cancel. Users instead can use the default top-right "X" to dismiss.


pleek91 (approved):
@aaazzam I looked at how to do this and this idea did not cross my mind. Super simple! This will remove the footer for any existing modals that don't have actions, but that should be fine yeah?

PR #1215, opened 2024-04-17
Fix vite cjs warning

pleek91:
# Description
This gets rid of the annoying vite cjs warning every time you run serve or build. 

~~Need to do some additional testing to verify this doesn't produce a broken build. But server seems to work fine.~~ Seems to work fine linking to a local build. 

https://vitejs.dev/guide/migration.html#deprecate-cjs-node-api


collincchoy (approved):
we should switch the package to ESM at some point

PR #1217, opened 2024-04-18
Refactor Sentiment Tokens

aaazzam:
Today we have four sentiment tokens: positive, negative, neutral, and warning. 

- These tokens take the same value on light and dark mode, so they've been moved to the true :root.  
- These colors were each assigned individual CSS Vars for their hue, sturation, and lightness indivudually, remixed into a hsl-token, and finally remixed into a color creating 4 * 5 * 2 = 40 CSS variables. 
    - This PR reduces the number of variables to 8 CSS variables. 

First of a handful of PRs to reduce the surface area of our design tokens to make them easier to maintain.


src/styles/index.css
```diff
@@ -20,9 +20,17 @@
   :root {
     --p-spacing-focus-ring: 2px;
     --p-spacing-focus-ring-offset: 1px;
-
     --p-radius-default: 6px;
     --p-scrollbar-width: 12px;
+    --destructive: 4 74% 49%;
+    --positive: 142 71% 45%;
+    --neutral: 240 5% 46%;
+    --warning: 40 92% 50%;
```
pleek91, line 28: Generally i'd suggest that global tokens have some sort of prefix/namespace. That way we don't run the risk (or we have less risk) or them being overridden by another library or a 3rd party extension and also adds css to the root.

pleek91: We only need these for the tailwind plugin yeah? Otherwise we could just put these values in the tokens below?

pleek91 (approved):
One though but probably unlikely to become an issue.

pleek91 (approved):
❤️

PR #1218, opened 2024-04-18
Enhancement: Add generics and better slots to PTabs

pleek91:
# Description
Wanted to fix a few issues. 
1. Tabs could not be readonly
2. Events did not reflect the tabs passed in

Updating these components to use generics and defineSlots fixes these

Note: There's a weird type issue in the install method for PTabs. I've run into this before and it seems to be related to using defineSlots producing some type other than what app.component wants. But everything seems to work as I'd expect when testing this.


src/components/Accordion/index.ts
```diff
@@ -2,9 +2,6 @@ import { App } from 'vue'
 import PAccordion from '@/components/Accordion/PAccordion.vue'
 
 const install = (app: App): void => {
-  // adding defineSlots doesn't match what app.component expects
-  // eslint-disable-next-line @typescript-eslint/ban-ts-comment, @typescript-eslint/prefer-ts-expect-error
-  // @ts-ignore
```
pleek91, line 7: Interestingly enough, this component doesn't even use defineSlots...

src/components/Tabs/PTabSelect.vue
pleek91, line 1: This components slots were inconsistent with the root PTabs component. Which likely would lead to implementation errors/bugs at some point. Bringing it inline (yay slot types!)

znicholasbrown (approved):

PR #1219, opened 2024-04-18
Standardize Background Tokens with color-mix

aaazzam:
This PR standardizes our background palette to be a single line in HSL space, e.g. "there's a base hue/saturation and other shades are just white/black mixins with it". Previously our color palette were more or less small perturbations around an HSL line. This doesn't change our application visually, yet -- but sets us up to do so much more easily. 

Note this makes use of the CSS function color-mix, which is supported in all major browsers since March 2023. 

Screenshot 2024-04-18 at 5 26 07 PM
Screenshot 2024-04-18 at 5 26 46 PM
Screenshot 2024-04-18 at 5 30 22 PM
Screenshot 2024-04-18 at 5 29 52 PM


pleek91 (approved):
love some good css

PR #1229, opened 2024-04-22
Spike/Automated releases using shell script

jimid27:
- Relates: https://github.com/PrefectHQ/platform/issues/6509
- Migrates to a workflow_dispatch release workflow vs a main merge workflow
- Uses an inputed release type (i.e patch, minor, major, etc) to then automatically bump the version using npm version <release_type>
- Create a simple bash script to allow for the release pipeline to be triggered from a local machine

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,14 +1,14 @@
 name: Release
 
 on:
-  push:
-    paths: 
-      - package.json
-    branches:
-      - main
-      
-jobs:
+  workflow_dispatch:
+    inputs:
+      release_type:
+        description: 'Type of release to perform'
+        required: true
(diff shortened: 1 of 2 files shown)


pleek91 (approved):

.github/workflows/release.yml
```diff
@@ -1,14 +1,14 @@
...
-    paths: 
-      - package.json
-    branches:
-      - main
-      
-jobs:
+  workflow_dispatch:
+    inputs:
+      release_type:
+        description: 'Type of release to perform'
```
jamiezieziula, line 7: don't think we need quotes here

jamiezieziula (approved):

PR #1233, opened 2024-04-23
Remove atomic classes for PButton

aaazzam:
We use a ton of 

```css
.p-button:not(:disabled):hover {
  background-color: var(--p-color-button-default-bg-hover);
  border-color: var(--p-color-button-default-border-hover);
  color: var(--p-color-button-default-text-hover);
}
```

when, say, 

```
enabled:hover:border-[color:var(--p-color-text-subdued)]
```

will do. This PR also removes conditional styles that, after a long rabbit hunt, were shown to not be doing anything.

### DIFF
--- a/src/components/Button/PButton.vue
+++ b/src/components/Button/PButton.vue
@@ -2,12 +2,13 @@
   <component
     :is="component"
     ref="el"
-    class="p-button"
+    class="group p-button"
     :class="classes"
     :disabled="disabled || loading"
+    :data-loading="loading"
     v-bind="componentProps"
   >
-    <div class="p-button__content">
+    <div class="p-button__content" :data-loading="loading">
       <template v-if="icon">
         <PIcon :size="iconSize" :icon="icon" class="p-button__icon" />
       </template>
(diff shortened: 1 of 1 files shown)


src/components/Button/PButton.vue
```diff
@@ -99,35 +100,36 @@
...
-  border-color: var(--p-color-button-default-border-active);
-  color: var(--p-color-button-default-text-active);
+.p-button__content { @apply
+  gap-1
+  inline-flex
+  flex-grow
+  justify-center
+  items-center
+  font-normal
+  group-data-[loading='true']:opacity-40
```
pleek91, line 132: is this like a `has` selector? Also, is `group` something special in tailwind?

pleek91: I see, it generates `.group[data-loading=true] .p-button__content`. Is `group` a special name in tailwind? Or can it be any class? Struggling to find good docs on that. 

My concern here is if we use `group` on say `p-form` and also add a loading state using this same technique. I believe all the buttons within the form would start spinning. If the we used a modifier class like `.p-button__content--loading` or if we can use a custom class other than `group` (like `p-button-data[loading='true']:opacity-40`) 

If I'm not understanding this right please help lol.

PR #1236, opened 2024-04-24
Enhancement: `PTable` 

znicholasbrown:
This PR makes several improvements to `PTable`, including:
- adding a border to that matches the row divider
- removing the `p-background` class from `PTableHead`
- tightening row and header vertical padding
- expanding column definitions to include min and max attributes; these help us better enforce column width definitions and prevent table blowouts
- truncating headers and cells
- adding title attributes to headers and cells
- swapping `font-semibold` for `font-bold` in header labels (semibold text doesn't provide a very clear distinction between headers and content)

Before:
Screenshot 2024-04-24 at 12 09 48 PM

After:
Screenshot 2024-04-24 at 12 09 45 PM

### DIFF
--- a/src/components/Table/PTable.vue
+++ b/src/components/Table/PTable.vue
@@ -31,7 +31,7 @@
               </template>
 
               <template v-for="(column, columnIndex) in visibleColumns" :key="column">
-                <PTableData :class="getColumnClasses(column, getValue(row, column.property), columnIndex, row, rowIndex)">
+                <PTableData :class="getColumnClasses(column, getValue(row, column.property), columnIndex, row, rowIndex)" :title="getValue(row, column.property)">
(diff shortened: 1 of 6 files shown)


src/components/Table/PTable.vue
```diff
@@ -12,7 +12,7 @@
               </template>
 
               <template v-for="column in visibleColumns" :key="column">
-                <PTableHeader :style="getColumnStyle(column)">
+                <PTableHeader :style="getColumnStyle(column)" :title="column.label">
```
znicholasbrown, line 15: Because of the way these slots are implemented, these title attributes will always work off the passed label and won't display any custom content - I think this is fine but open to pushback if that's not a clear devex.

pleek91 (approved):

PR #1240, opened 2024-04-24
Button Design + Compat-layer

aaazzam:
This PR is a backwards-compatible compat-layer between our old taxonomy of button variants and sizes and new:

- default, primary, flat, and dangerous 
- outline, default, ghost, destructive, and link (new, unimplemented).   

This is backwards compatible in the sense that all old props still work, all existing buttons do the same thing.


pleek91 (approved):
So ghost is the new name for flat?

PR #1250, opened 2024-04-25
BEM managed by CVA in PButton

aaazzam:
This PR moves PButton to adopt CVA to manage it's BEM hierarchy. 

- While we're in the ugly haircut phase, a compat layer between old and new props are maintained. 

Along the way, the styles for `selected` are properly handled via aria-selected. 

This is likely the penultimate PR, which will free us from the ~80 CSS Vars to style buttons.


znicholasbrown:
Discussed this sync - there's a bit of cleanup to do here but largely this is the appropriate direction, we'll work out some of the hairier details in the next few days.

znicholasbrown (approved):

src/components/Button/PButton.vue
```diff
@@ -23,16 +23,54 @@
...
+    },
+  )
+
+  type ButtonProps = VariantProps<typeof button>
+
   const props = defineProps<{
-    variant?: 'default' | 'secondary' | 'destructive' | 'outline' | 'ghost' | 'link',
-    size?: 'default' | 'sm' | 'lg' | 'icon',
+    variant?: ButtonProps['variant'],
+    size?: ButtonProps['size'],
```
pleek91, line 73: ❤️

PR #1255, opened 2024-04-30
Radix PAccordion 

aaazzam:
We have three uses of PAccordion, of which one uses selected?, a parameter I'd like to deprecate. Note that current uses of PAccordion don't take advantage of custom header or content slots, instead passing simple data straight to the content slot. That API surface is also removed. 

I don't know off the top of my head how to stage all these changes between libraries to not break anything. 🤔

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "highlight.js": "11.9.0",
     "lodash.debounce": "4.0.8",
     "marked": "4.3.0",
-    "radix-vue": "^1.7.3",
+    "radix-vue": "1.7.3",
     "tailwind-merge": "^2.3.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "highlight.js": "11.9.0",
     "lodash.debounce": "4.0.8",
     "marked": "4.3.0",
+    "radix-vue": "^1.7.3",
     "tailwind-merge": "^2.3.0"
   }
 }
--- a/public/tailwindPlugin.js
+++ b/public/tailwindPlugin.js
@@ -8,7 +8,12 @@ const borderRadius = {
   default: 'var(--p-radius-default)',
 }
 
+const borderColor = {
+  DEFAULT: 'hsl(var(--p-border) / <alpha-value>)',
+}
+
 const colors = {
+  'border': 'hsl(var(--p-destructive) / <alpha-value>)',
   'input': 'hsl(var(--p-input) / <alpha-value>)',
   'background': 'hsl(var(--p-background) / <alpha-value>)',
   'primary': 'hsl(var(--p-primary) / <alpha-value>)',
@@ -45,13 +50,33 @@ const fontFamily = {
   mono: ['InconsolataVariable', ...defaultTheme.fontFamily.mono],
 }
 
+const keyframes = {
+  'accordion-down': {
+    from: { height: 0 },
+    to: { height: 'var(--radix-accordion-content-height)' },
+  },
+  'accordion-up': {
+    from: { height: 'var(--radix-accordion-content-height)' },
+    to: { height: 0 },
+  },
+}
+
+const animation = {
+  'accordion-down': 'accordion-down 0.2s ease-out',
+  'accordion-up': 'accordion-up 0.2s ease-out',
+}
+
+
 const extend = {
   borderRadius,
+  borderColor,
   colors,
   spacing,
   ringWidth,
   ringOffsetWidth,
   fontFamily,
+  keyframes,
+  animation,
 }
 
 const darkMode = 'class'
--- a/src/components/Accordion/PAccordion.vue
+++ b/src/components/Accordion/PAccordion.vue
@@ -1,6 +1,6 @@
 
 <template>
-  <AccordionRoot v-bind="forwarded" type="single" :collapsible="true">
+  <AccordionRoot v-bind="forwarded">
     <AccordionItem v-for="section in sections" :key="section" :value="section">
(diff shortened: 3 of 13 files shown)


src/components/Accordion/PAccordionContent.vue
```diff
@@ -0,0 +1,35 @@
...
+<style scoped>
+.p-accordion-content{
+  @apply
+    overflow-hidden
+    text-sm
+    transition-all
+    data-[state=closed]:animate-accordion-up
+    data-[state=open]:animate-accordion-down
+}
+</style>
```
znicholasbrown, line 35:
```suggestion
<style>
.p-accordion-content { @apply
    overflow-hidden
    text-sm
    transition-all
    data-[state=closed]:animate-accordion-up
    data-[state=open]:animate-accordion-down
}
</style>
```
we need a linter for these

src/components/Accordion/PAccordion.vue
```diff
@@ -1,144 +1,36 @@
...
-                <slot name="heading" :section="section" :selected="isSelected(section)">
-                  {{ section }}
-                </slot>
-              </slot>
-            </span>
-            <slot name="icon" :section="section">
-              <p-icon size="small" icon="ChevronDownIcon" class="p-accordion__icon" :class="getIconClass(section)" />
-            </slot>
-          </button>
+  <AccordionRoot v-bind="forwarded" type="single" :collapsible="true">
```
znicholasbrown, line 3: Why not expose `type` and `collapsible` as props for callers to choose?

public/tailwindPlugin.js
```diff
@@ -45,13 +50,33 @@ const fontFamily = {
...
 
+const keyframes = {
+  'accordion-down': {
+    from: { height: 0 },
+    to: { height: 'var(--radix-accordion-content-height)' },
+  },
+  'accordion-up': {
+    from: { height: 'var(--radix-accordion-content-height)' },
+    to: { height: 0 },
+  },
```
znicholasbrown, line 61: Is this best-practice? I'd expect since these are inverse animations to be able to reverse the animation direction and get the same effect with fewer lines of code

package.json
```diff
@@ -78,6 +78,7 @@
     "highlight.js": "11.9.0",
     "lodash.debounce": "4.0.8",
     "marked": "4.3.0",
+    "radix-vue": "^1.7.3",
```
znicholasbrown, line 81: I think it'd be a good idea to pin the version here so this doesn't bite us

src/components/Accordion/PAccordionContent.vue
```diff
@@ -0,0 +1,35 @@
+<template>
+  <AccordionContent
+    v-bind="delegatedProps"
+    class="p-accordion-content"
+  >
+    <div class="pb-4 pt-0" :class="props.class">
+      <slot />
+    </div>
```
znicholasbrown, line 8: Why an interior div instead of something like:
```vue
<template>
  <AccordionContent
    v-bind="delegatedProps"
    class="p-accordion-content"
  >
  <slot />
</AccordionContent>

...

<style>
.p-accordion-content { @apply
  pb-4
  pt-0
}
</style>
```

Generally using an arbitrary inner element like this makes the component less flexible than intended and more difficult to style, whether because you need to pass classes to the inner div like you're doing here OR you need to pass the to the outer one and can't.

src/components/Accordion/PAccordionTrigger.vue
```diff
@@ -0,0 +1,60 @@
...
+
+.p-accordion-trigger-icon {
+  @apply
+    h-4
+    w-4
+    shrink-0
+    transition-transform
+    duration-200
+}
+</style>
```
znicholasbrown, line 60:
```suggestion
<style scoped>
.p-accordion-trigger { @apply
    cursor-pointer
    flex
    flex-1
    items-center
    justify-between
    py-4
    font-medium
    transition-all
    hover:underline
    [&[data-state=open]>svg]:rotate-180
}

.p-accordion-trigger-icon { @apply
    h-4
    w-4
    shrink-0
    transition-transform
    duration-200
}
</style>
```

linting; also that data attribute svg selector is a little whack; can you convert that to a hopefully more readable class instead? something like:

```
.p-accordion-trigger[data-state=open] > svg { @apply
  rotate-180
}
```
i think that'll be a lot clearer to anyone reading this css

znicholasbrown (commented):
A few comments @aaazzam, mostly this is looking/feeling good to me

src/components/Accordion/PAccordionTrigger.vue
pleek91: +1 that feels like codegolf

src/components/Accordion/PAccordion.vue
```diff
@@ -1,144 +1,40 @@
...
-  }
-
-  function getHeadingId(section: T): string {
-    return `${kebabCase(section)}-heading`
-  }
-
-  function getContentId(section: T): string {
-    return `${kebabCase(section)}-content`
-  }
+  const forwarded = useForwardPropsEmits(props, emits)
```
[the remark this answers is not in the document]
pleek91: Resolved sync. This is really only needed for emits. And it's because when you use `defineEmits` that removes any matching event handlers from attrs similar to how any props you define are removed from attrs. So this composition gives them back to you in the same `'onUpdate:modelValue': f()` format that vue expects when you bind an event handler without the `@`.

pleek91 (approved):

znicholasbrown (approved):

PR #1256, opened 2024-04-30
PButtonGroup: Add slot type and add support for generics and readonly options

pleek91:
# Description
Small quality of life improvements.


src/components/ButtonGroup/.eslintrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "@typescript-eslint/no-unnecessary-type-arguments": "off"
```
pleek91, line 3: This rule is annoying and I think we should remove it. But didn't want to get to distracted

znicholasbrown (approved):

PR #1257, opened 2024-04-30
Radix PCard

aaazzam:
Adds Radix PCard components, exchanges flat prop for "ghost" variant.

### DIFF
--- a/src/components/Accordion/PAccordionItem.vue
+++ b/src/components/Accordion/PAccordionItem.vue
@@ -1,8 +1,9 @@
 <template>
   <AccordionItem
-    v-bind="forwardedProps"
+    v-bind="props"
     class="border-b"
-    :class="props.class"
+    :class="attrClasses"
+    :style="attrStyles"
   >
     <slot />
   </AccordionItem>
@@ -10,15 +11,11 @@
 
 <script setup lang="ts">
   import { AccordionItem, type AccordionItemProps, useForwardProps } from 'radix-vue'
-  import { type HTMLAttributes, computed } from 'vue'
+  import { useAttrsStylesAndClasses } from '@/compositions'
 
-  const props = defineProps<AccordionItemProps & { class?: HTMLAttributes['class'] }>()
+  const { classes: attrClasses, styles: attrStyles } = useAttrsStylesAndClasses()
 
-  const delegatedProps = computed(() => {
-    // eslint-disable-next-line id-length, no-unused-vars
-    const { class: _, ...delegated } = props
-    return delegated
-  })
 
-  const forwardedProps = useForwardProps(delegatedProps)
+  // eslint-disable-next-line vue/no-unused-properties
+  const props = defineProps<AccordionItemProps>()
 </script>
--- a/src/components/Accordion/PAccordionTrigger.vue
+++ b/src/components/Accordion/PAccordionTrigger.vue
@@ -1,9 +1,10 @@
 <template>
   <AccordionHeader class="flex">
     <AccordionTrigger
-      v-bind="delegatedProps"
+      v-bind="props"
       class="p-accordion-trigger"
-      :class="props.class"
+      :style="attrStyles"
+      :class="attrClasses"
     >
       <slot />
       <slot name="icon">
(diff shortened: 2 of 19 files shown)


src/components/Card/PCard.vue
```diff
@@ -1,32 +1,36 @@
...
 
   const props = defineProps<{
-    flat?: boolean,
+    class?: HTMLAttributes['class'],
+    variant?: CardProps['variant'],
   }>()
-
-  const classes = computed(() => ({
-    'p-card--flat': props.flat,
-    'p-background': !props.flat,
```
pleek91, line 16: How are you accounting for nested cards? Currently we have a system with p-background so that if you end up with a card inside a card they are not the same color. Obviously this isn't a good ux pattern and we don't do it intentionally. But user generated schema based forms can have this. 

I see now the card has a border around it. Is that sufficient for this?

src/components/Card/PCard.vue
```diff
@@ -1,32 +1,36 @@
...
+  import CardBase, { type CardProps } from '@/components/Card/PCardBase.vue'
+  import CardContent from '@/components/Card/PCardContent.vue'
+  import CardDescription from '@/components/Card/PCardDescription.vue'
+  import CardFooter from '@/components/Card/PCardFooter.vue'
+  import CardHeader from '@/components/Card/PCardHeader.vue'
+  import CardTitle from '@/components/Card/PCardTitle.vue'
 
   const props = defineProps<{
-    flat?: boolean,
+    class?: HTMLAttributes['class'],
```
pleek91, line 33: You shouldn't make class a prop. Vue will automatically merge any classes you bind here with the classes added at implementation. This applies to all of these components.

src/components/Card/PCardBase.vue
```diff
@@ -0,0 +1,52 @@
+<template>
+  <div
+    class="p-card"
+    :class="normalizeClass([props.class, card({ variant })])"
```
pleek91, line 4: again, don't make a class prop. But also if you have an array of classes, just bind them directly to the class attribute, vue will handle the normalizing for you. So you don't need to use normalizeClass here.

src/components/Card/PCardContent.vue
pleek91, line 1: This applies to most of these. Why is this a component and not just part of the PCard component? This is basically just an abstraction of a div and a class.

src/components/Card/PCardBase.vue
```diff
@@ -0,0 +1,52 @@
...
+  export type CardProps = VariantProps<typeof card>
+
+  const props = defineProps<{
+    class?: HTMLAttributes['class'],
+    variant?: CardProps['variant'],
+  }>()
+</script>
+
+
+<style scoped>
```
pleek91, line 36: if we're using BEM classes we shouldn't need to use `scoped`. There is overhead associated with that and it increases specificity. So we should probably remove the scoped attribute from all of these style blocks.

src/components/Accordion/PAccordionTrigger.vue
```diff
@@ -0,0 +1,60 @@
...
+  import { type HTMLAttributes, computed } from 'vue'
+
+  const props = defineProps<AccordionTriggerProps & { class?: HTMLAttributes['class'] }>()
+
+  const delegatedProps = computed(() => {
+    // eslint-disable-next-line id-length, no-unused-vars
+    const { class: _, ...delegated } = props
+
+    return delegated
+  })
```
pleek91, line 34: You should use the [`useAttrsStylesAndClasses` ](https://github.com/PrefectHQ/prefect-design/blob/main/src/compositions/attributes.ts#L14)composition for this. 

That'll give you the classes, styles, and attrs as separate reactive refs so you can bind classes and styles to the root and attrs to the child.

pleek91: You might want to set `inheritAttrs: false` too if you don't want any attrs to fallthrough to the root element.

src/components/Accordion/PAccordionItem.vue
```diff
@@ -0,0 +1,24 @@
...
+
+  const props = defineProps<AccordionItemProps & { class?: HTMLAttributes['class'] }>()
+
+  const delegatedProps = computed(() => {
+    // eslint-disable-next-line id-length, no-unused-vars
+    const { class: _, ...delegated } = props
+    return delegated
+  })
+
+  const forwardedProps = useForwardProps(delegatedProps)
```
pleek91, line 23: Why do we want to skip vue's default boolean casting?

pleek91:
The accordion looks nice and the keyboard support ❤️

src/components/Card/PCard.vue
```diff
@@ -1,32 +1,33 @@
 <template>
-  <div class="p-card" :class="classes">
-    <slot />
-  </div>
+  <CardBase
+    :variant="props.variant"
+  >
+    <CardHeader v-if="$slots.header || $slots.title || $slots.description">
```
pleek91, line 5: this is checking for slots.header but it doesn't appear there is a header slot?

pleek91 (approved):
One extra slot I think but otherwise LGTM

pleek91:
I feel like this branch needs rebased on main or something. Diff is weird considering the accordion changes were merged into main already

PR #1263, opened 2024-05-02
PDialog Radix

aaazzam:
PR to change Dialog logic / styling to Radix. 

Note this does not emit a close event, but this has been handled in the one application of PDialog.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "lodash.debounce": "4.0.8",
     "marked": "4.3.0",
     "radix-vue": "1.7.3",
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -79,6 +79,7 @@
     "lodash.debounce": "4.0.8",
     "marked": "4.3.0",
     "radix-vue": "1.7.3",
-    "tailwind-merge": "^2.3.0"
+    "tailwind-merge": "^2.3.0",
```
pleek91, line 82: Aren't we not using tailwind merge?

pleek91 (approved):

PR #1273, opened 2024-05-03
Chore: Tailwind config preset and TS migration

znicholasbrown:
While a chore, this PR includes some notable changes:
1. The package now expliticly exports all files instead of just those in `/dist` - this is a convenience for downstream consumers to allow better file discovery/not limit to exports in `/dist`
2. Postcss config has been migrated to the modern `.cjs` syntax
3. Tailwind plugin has been removed in favor of `presets` syntax (a plugin still exists for adding utilities but this doesn't need to be consumed by downstream applications anymore)
4. Tailwind config has been migrated to ES modules
5. Standardizes tsconfig with other packages in our ecosystem; a future enhancement would be to centralize and extend this config but that felt out of scope of this PR.


package.json
```diff
@@ -32,12 +32,10 @@
     ".": {
       "import": "./dist/prefect-design.mjs",
       "require": "./dist/prefect-design.umd.js"
-    },
-    "./dist/style.css": "./dist/style.css",
```
znicholasbrown, line 36: Downstream applications will need to update their reference to this file

pleek91 (approved):

PR #1292, opened 2024-05-14
Enhancement: `PTable` `row-after` slot

znicholasbrown:
This PR adds a new slot, `row-after` that allows callers to specify any sort of additional content below the content of a row. 

Demo of this in action:

https://github.com/PrefectHQ/prefect-design/assets/27291717/4c0ba1de-a1fb-46a7-aa4a-c2cee8365528

### DIFF
--- a/demo/sections/components/Table.vue
+++ b/demo/sections/components/Table.vue
@@ -97,8 +97,8 @@
 
         <template #row-after="{ row }">
           <PTableRow v-if="isExpanded(row)">
-            <PTableData :colspan="columns.length">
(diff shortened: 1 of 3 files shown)


pleek91 (approved):
Is there a symantic way html wise for this? Curious if there's a more accessible way than just inserting extra tr elements in between rows.

znicholasbrown:
Great question - the best example I could find for this was the [tanstack table expanding rows](https://tanstack.com/table/v8/docs/framework/react/examples/expanding) demo, which compiles down to exactly this (and is what I modeled it after) - the truth is I'm not sure if there's a semantic way to handle this that doesn't break cell conventions so this felt like the next best thing

PR #1301, opened 2024-05-20
[p-table] Remove built-in titles on p-table data cells.

collincchoy:
Not sure that these titles add much. If anything, non-primitive values take a hit here

image

### DIFF
--- a/src/components/Table/PTable.vue
+++ b/src/components/Table/PTable.vue
@@ -31,7 +31,7 @@
               </template>
 
               <template v-for="(column, columnIndex) in visibleColumns" :key="column">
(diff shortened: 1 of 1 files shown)


znicholasbrown (approved):
yeah this is def the right call

pleek91:
I think this was probably here because of truncation. Any concerns about that? I agree this probably isn't super useful

collincchoy:
> I think this was probably here because of truncation. Any concerns about that? I agree this probably isn't super useful

Hm did a quick audit across the app, I think most tables scroll by default (at the table width level). Probably added for the deployments list/table... where the names are truncated but I think it probably makes more sense to handle that there rather than here. 

(I also noticed a couple of other oddities w/ title as i perused some other tables. things like the title shows the UUID id for what's otherwise an already human-readable value. possibly useful but imo looks unintentional)

PR #1303, opened 2024-05-21
Chore: Font dependencies

znicholasbrown:
This PR updates the fontsource packages and makes them dependencies instead of dev-dependencies. I'm hoping this change will let us compile stylesheets from source in downstream packages, since the blocker at the moment is the missing dependency.

### DIFF
--- a/src/tailwind.config.ts
+++ b/src/tailwind.config.ts
@@ -120,11 +120,7 @@ const plugins = [
 ]
 
 export default {
-  content: [
-    './src/**/*.{vue,js,ts}',
-    './demo/index.html',
-    './demo/**/*.vue',
-  ],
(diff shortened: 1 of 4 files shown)


src/tailwind.config.ts
```diff
@@ -120,11 +120,7 @@ const plugins = [
 ]
 
 export default {
-  content: [
-    './src/**/*.{vue,js,ts}',
-    './demo/index.html',
-    './demo/**/*.vue',
-  ],
+  content: ['./src/**/*.{vue}'],
```
znicholasbrown, line 123: We're already doing this in the top-level block so this one probably just needs the vue src glob?

pleek91 (approved):
Nice, glad to update these. Hopefully this helps

PR #1308, opened 2024-05-22
Chore: Update tailwind with the correct font family names

znicholasbrown:
Oops


pleek91 (approved):
We shall never speak of this again

PR #1314, opened 2024-05-24
Fix PDateInput closing when picking a month or year

pleek91:
# Description
When picking a month or year from the date picker the popover would close, preventing the user from actually selecting their desired date. This was happening because the element clicked on was no longer part of the DOM when PPopOver did its check to auto close.

### DIFF
--- a/src/components/PopOver/PPopOver.vue
+++ b/src/components/PopOver/PPopOver.vue
@@ -23,7 +23,7 @@
   import { useMostVisiblePositionStyles } from '@/compositions/position'
   import { usePopOverGroup } from '@/compositions/usePopOverGroup'
   import { PositionMethod } from '@/types/position'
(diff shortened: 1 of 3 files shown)


src/components/PopOver/PPopOver.vue
```diff
@@ -23,7 +23,7 @@
   import { useMostVisiblePositionStyles } from '@/compositions/position'
   import { usePopOverGroup } from '@/compositions/usePopOverGroup'
   import { PositionMethod } from '@/types/position'
-  import { randomId } from '@/utilities'
+  import { isDefined, randomId } from '@/utilities'
```
pleek91:
```suggestion
  import { randomId } from '@/utilities'
```

znicholasbrown (approved):
lgtm

PR #1324, opened 2024-05-30
Enhancement: `PModal` prop to hide close button

znicholasbrown:
Also updates the default close button text to "Close" instead of "Cancel" since that didn't make much sense and we're overriding that behavior almost every time we use it.

### DIFF
--- a/src/components/Modal/PModal.vue
+++ b/src/components/Modal/PModal.vue
@@ -25,7 +25,14 @@
                   <span class="p-modal__title">{{ title }}</span>
                 </slot>
               </div>
-              <p-button class="p-modal__x-button" icon="XMarkIcon" small flat @click="modalScope.close" />
+              <p-button
+                v-if="!hideCloseButton"
+                class="p-modal__x-button"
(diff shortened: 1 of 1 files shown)


src/components/Modal/PModal.vue
```diff
@@ -34,9 +41,9 @@
 
             <div v-if="$slots.actions" class="p-modal__footer">
               <slot name="actions" v-bind="modalScope" />
-              <slot name="cancel" v-bind="modalScope">
+              <slot v-if="!hideCloseButton" name="cancel" v-bind="modalScope">
```
dylanbhughes, line 44: I wish there were a way to alias the slot name so we could rename the slot from `cancel -> close` without breaking changes

dylanbhughes (approved):

src/components/Modal/PModal.vue
znicholasbrown: Agreed, I decided that was more than I wanted to tackle right now 😞

PR #1344, opened 2024-06-10
Enhancement: `PWizard` step header UX

znicholasbrown:
This PR removes the step header `cursor: pointer` for non-interactive header elements. When in a linear form this means that future steps won't show the cursor unless the steps before them have passed validation.

Resolves: https://github.com/PrefectHQ/nebula-ui/issues/4800


zhen0:
LGTM!  Thanks Nicholas

pleek91 (approved):

PR #1381, opened 2024-07-13
BugFix: incorrect css custom property syntax

stackoverfloweth:
Only when using `light` mode, value for `--p-color-button-primary-bg` was set to `hsl(--p-primary)`. I'm guessing this is intended to be using the custom property `p-primary` but is missing the `var()` syntax.


collincchoy (approved):
🙈 - Thank you!

PR #1405, opened 2024-08-12
Fix module declaration

pleek91:
# Description
This release replaces declare module `@vue/runtime-core` with declare module `vue` like it's [supposed to be](https://vuejs.org/guide/extras/web-components.html#web-components-and-typescript).

### DIFF
--- a/src/components/index.ts
+++ b/src/components/index.ts
@@ -327,7 +327,7 @@ export const installs = [
   installPWizard,
 ]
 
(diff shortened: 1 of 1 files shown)


collincchoy (approved):
trying to remember exactly but i think i did this also during troubleshooting for https://github.com/PrefectHQ/prefect-design/pull/1370. 

which... is to say iirc - cool.

PR #1452, opened 2024-09-23
Enhancement: Make tooltip collision avoidance opt-out

znicholasbrown:
This should be the default behavior; also makes the mono xs font the default for content (we were overriding this downstream enough)


pleek91 (approved):
Yes, I was going to make this same PR

PR #1459, opened 2024-10-02
Add a size vmodel for the resizable section

pleek91:
# Description
Adds an optional `size` v-model that can be used to set the initial size of the layout's aside when mounted. It can also be used as a v-model. The component will still keep track of the current size internally without the size prop. 

Note:
The `size` prop is respected over the `--p-layout-resizable-aside-size` custom property which sets a default value for if the size prop is missing.


znicholasbrown (approved):

znicholasbrown:
Nice solution 👌

PR #1474, opened 2024-10-17
Bugfix: PButton sizing issues

znicholasbrown:
Fixes some sizing issues with small icon-only buttons as well as an aria attribute issue that was causing errors with `PSelectButton`

Before: ( 28px height ); in order this is a `p-select`, `p-icon-button-menu`, `p-button` (text), `p-button` (icon)
Screenshot 2024-10-17 at 12 15 19 PM

After ( 30px height, same as `PSelect`)
Screenshot 2024-10-17 at 12 15 02 PM

### DIFF
--- a/src/components/Button/PButton.vue
+++ b/src/components/Button/PButton.vue
@@ -9,11 +9,11 @@
   >
     <div class="p-button__content">
       <template v-if="icon">
-        <PIcon :size="iconSize" :icon :solid class="p-button__icon" />
+        <PIcon :size="iconSize ?? size" :icon :solid class="p-button__icon" />
       </template>
       <slot />
       <template v-if="iconAppend">
-        <PIcon :size="iconSize" :icon="iconAppend" :solid="solidAppend ?? solid" class="p-button__icon" />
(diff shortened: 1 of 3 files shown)


src/components/Icon/PIcon.vue
```diff
@@ -13,7 +13,7 @@
   const props = defineProps<{
     icon: Icon,
     solid?: boolean,
-    size?: 'small' | 'default' | 'large',
+    size?: 'small' | 'default' | 'large' | 'sm' | 'lg',
```
znicholasbrown, line 16: This is a bridge change to allow us to use size the same way we do for buttons

src/components/Button/PButton.vue
```diff
@@ -234,6 +234,11 @@
   font-normal
 }
 
+.p-button__icon { @apply
+  shrink-0
+  grow
+}
+
```
znicholasbrown, line 241: Icons were shrinking cause of the flex

pleek91 (approved):

PR #1488, opened 2024-10-28
Add the ability to scroll an item into view in the virtual scroller

pleek91:
# Description
Adds a new utility `getVirtualScroller` where you can get access to a virtual scroller instance by its given name (which is now a prop on the virtual scroller component). This way we can grab a scroller at any point in the component tree. The available actions are
- `makeItemVisible` - Accepts an itemKey and forces the chunk that item belongs to to be visible
- `scrollItemIntoView` - Accepts an itemKey and automatically scrolls to that item.

### DIFF
--- a/src/components/VirtualScroller/PVirtualScroller.vue
+++ b/src/components/VirtualScroller/PVirtualScroller.vue
@@ -1,9 +1,9 @@
 <template>
-  <component :is="element" class="p-virtual-scroller">
+  <component :is="element" ref="scrollerRef" class="p-virtual-scroller">
     <template v-for="(chunk, chunkIndex) in chunks" :key="chunkIndex">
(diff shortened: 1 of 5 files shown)


src/components/VirtualScroller/PVirtualScroller.vue
```diff
@@ -64,11 +139,21 @@
     return props.chunkSize * chunkIndex + itemChunkIndex
   }
 
-  watch(() => props.items, () => check(bottom))
+  function getChunkHeight(index: number): number {
+    return props.itemEstimateHeight * chunks.value[index].length
+  }
```
znicholasbrown, line 145: Oh wow, we were severely underestimating these heights huh

znicholasbrown (approved):
lgtm!

PR #1494, opened 2024-10-29
Bugfix: Don't create dates to calculate intervals for `PDateRangeSelect`

znicholasbrown:
The upcoming DST switchover surfaced a bug in the way we're calculating the span labels for `PDateRangeSelect`. This results in an off-by-DST-interval (usually 1 hour but not everywhere!) label whenever the encompassing timeframe includes a DST boundary.

This PR addresses this by removing dates from the equation, instead relying on seconds to a given timeframe. To avoid needing some much more complicated calculations around DST boundaries, month seconds, and leap years, I've eliminated month and year support, with weeks as the largest interval this picker supports for now.

### DIFF
--- a/demo/sections/components/DateRangeSelect.vue
+++ b/demo/sections/components/DateRangeSelect.vue
@@ -35,7 +35,7 @@
   const exampleState = ref<State>()
   const disabled = ref(false)
 
-  const value = ref({"type":"span","seconds":-604800})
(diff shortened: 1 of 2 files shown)


pleek91 (approved):
Nice, that makes sense to not rely on `intervalToDuration` or `addSeconds` 👍

PR #1499, opened 2024-11-06
Enhancement: Add new icon for worker

dylanbhughes:


src/components/Icon/icons/p-worker.svg
pleek91, line 1: I think the viewbox for this icon is much smaller than the others. Probably want them to all be the same/similar

https://github.com/PrefectHQ/prefect-design/tree/main/src/components/Icon/icons

dylanbhughes: Ah I see, the viewbox for WorkPool is also `viewBox="0 0 24 24"`, let me get back with Karim

dylanbhughes: Screenshot 2024-11-06 at 12 30 34 PM
looks like a fair number are off

dylanbhughes: @pleek91 Looks like there was a stray `height` and `width` still in the SVG -- fixed!

pleek91 (approved):

src/components/Icon/icons/p-worker.svg
pleek91: ah that looks better now! ty

PR #1525, opened 2024-11-21
Enhancement: Allow customizing p-drawer teleport location

znicholasbrown:
Makes this a little easier to work with in local contexts


src/components/Drawer/PDrawer.vue
```diff
@@ -37,13 +37,15 @@
     open?: boolean,
     resizable?: boolean,
     placement?: DrawerPlacement,
+    to?: string,
```
pleek91, line 40: I'd destructure this and default it to body still

pleek91 (approved):

PR #1526, opened 2024-11-21
Enhancement: Drawer teleporting

znicholasbrown:
Adds a demo for disabled drawer teleports and generally cleans up the drawer component to use more modern syntax

### DIFF
--- a/src/components/Drawer/PDrawer.vue
+++ b/src/components/Drawer/PDrawer.vue
@@ -6,7 +6,6 @@
           <template #aside>
             <div class="p-drawer__aside" v-bind="attrs">
               <slot v-bind="drawerScope" />
-              {{ }}
             </div>
           </template>
 
--- a/src/components/Drawer/PDrawer.vue
+++ b/src/components/Drawer/PDrawer.vue
@@ -3,13 +3,14 @@
     This coalesce is important because Vue seems to evaluate the to prop for teleportation
     before it finishes props destructuring
   -->
-  <teleport :to="to ?? 'body'" :disabled="disableTeleport">
+  <teleport :to :disabled="disableTeleport">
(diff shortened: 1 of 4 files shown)


src/components/Drawer/PDrawer.vue
```diff
@@ -27,40 +31,29 @@
...
   import { PLayoutResizable } from '@/layouts'
   import { DrawerPlacement, keys } from '@/types'
   import { isKeyEvent } from '@/utilities'
 
-  const props = defineProps<{
-    open?: boolean,
+  const {
+    resizable = false,
+    placement = 'left',
+    to,
```
pleek91, line 43: nit: Then you don't need to null coalesce on the teleport element. 
```suggestion
    to = 'body',
```

pleek91 (approved):

src/components/Drawer/PDrawer.vue
znicholasbrown: you'd think that but this doesn't work for some reason

znicholasbrown: ohhhhhhh. it's cause the peer dependency for vue was too low

pleek91: oh, nevermind then. We should update that some time lol.

znicholasbrown: we're on 3.4.x and props destructuring wasn't introduced till 3.5

znicholasbrown: i'm just gonna update there here unless you have reservations?

pleek91: none, we're already using 3.5 in production so that's whats actually being used for this in apps already. Shouldn't change anything.

PR #1585, opened 2025-03-13
PDateRangeSelect - Omit year from label when it is the current year

pleek91:
# Description
The input's description of the selected date range can be very long. Omitting the year when its the current year makes it more concise and also a little easier to read. 

Before
image

After
image


znicholasbrown (approved):
love

PR #1633, opened 2025-06-11
Enhancement: `PMarkdownRenderer` anchors

znicholasbrown:
Removes the default anchor elements used in the markdown renderer in favor of a more straightforward heading element in most cases.


zhen0 (approved):
Could you remove the unused imports before merge?

(Approving anyway so as not to slow you down)

znicholasbrown:
ah yes, good catch, ty!

PR #1650, opened 2025-07-01
feat: Adds Past month option to date range

devinvillarosa:
Adds a Past month option

https://github.com/user-attachments/assets/84f3bd31-b4e7-4a53-876b-247b184b7e8b


src/components/DateRangeSelect/utilities.ts
```diff
@@ -39,8 +39,14 @@ export function getDateRangeSelectValueLabel(value: DateRangeSelectValue): strin
 function getDateSpanLabel({ seconds }: DateRangeSelectSpanValue): string {
   const absSeconds = Math.abs(seconds)
 
+  // nb: Because secondsInMonth is an average, just return Past month label when option is selected
```
devinvillarosa, line 42: Below is why we just return `Past month` instead of being precise

Screenshot 2025-07-01 at 8 39 44 AM

znicholasbrown (approved):

PR #1653, opened 2025-07-01
feat: Re-orders date range with new Past 7 and 30 day options

devinvillarosa:
Based on recent feedback, have the following options:

> Past hour
Past day
Past 7 days
Past 30 days

https://github.com/user-attachments/assets/cfd5fac1-c0e7-4293-8f38-673996a941ac

### DIFF
--- a/src/components/DateRangeSelect/PDateRangeSelectOptions.vue
+++ b/src/components/DateRangeSelect/PDateRangeSelectOptions.vue
@@ -3,9 +3,10 @@
 </template>
 
 <script lang="ts" setup>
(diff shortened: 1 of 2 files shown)


znicholasbrown (approved):
we don't like "past week" and "past month"?

devinvillarosa:
> we don't like "past week" and "past month"?

It doesn't seem as clear: https://linear.app/prefect/issue/ENG-2396/display-past-month-as-a-dropdown-option-on-the-global-time-filter#comment-2695f646

Which I kinda agree. Say if today is July 15th, `Past month` can be interpreted as `Jun 1 - Jun 30`, rather than the last 30 days

PR #1697, opened 2026-04-10
allow controls inside of PBaseInput to receive listeners as well

stackoverfloweth:
When `p-base-input` separates `attrs, listeners, styles, classes` from `useAttrs` it only ever sends the `attrs` through to the underlying #control slot element. This separation makes sense for things like class/style to prevent doubling up of class names or styles on both the parent element and the control. However, for listeners I think having them apply to both the parent and the control is a better design. 

With this PR, the control now has the option of consuming these events without having to explicitly "forward" listeners, which you're currently doing for `focus` but there are hundreds of native events across all components that build off `p-base-input`.

```ts
<!-- not possible without fix -->
<PTextInput @blur="handleBlurEvent"/>
```

### DIFF
--- a/src/components/BaseInput/PBaseInput.vue
+++ b/src/components/BaseInput/PBaseInput.vue
@@ -60,6 +60,7 @@
 
   const attrsForControl = computed(() => ({
     ...attrs.value,
+    ...listeners.value,
     disabled: props.disabled,
     class: {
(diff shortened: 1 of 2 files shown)


znicholasbrown (approved):
nice, thanks!

